PR #1, opened 2016-09-06
Spelling and standardization fixes to match GURPS Magic

dehnert:
There are several spells in Magic.spl that don't actually match the names used
in GURPS Magic. This changes them to match. I don't know if you necessarily want this -- for some of these I could believe this is a ~difference in your style guide, and even for the others I don't know if this breaks things for existing characters (looks like no?). Anyway, I was trying to merge information from the GCS spell list and another one and ran across these discrepancies, so figured I'd submit this. (Also, sorry if PR's are the wrong way to suggest changes. Since I had a patch, I figured this might be the easiest.)

Summary of changes:
- Change plural/non-plural status of spells
- "Rain Ice Daggers" is "Rain of Ice Daggers"
- "Remove Enchant" is "Remove Enchantment"
- "Steel Wraith" is "Steelwraith"
- "Rooted Feed" should be "Feet"

### DIFF
--- a/Library/Spells/Magic.spl
+++ b/Library/Spells/Magic.spl
@@ -1505,7 +1505,7 @@
 		</prereq_list>
 	</spell>
 	<spell version="2">
-		<name>Bless Plant</name>
+		<name>Bless Plants</name>
 		<college>Plant</college>
 		<power_source>Arcane</power_source>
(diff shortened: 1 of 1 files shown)


richardwilkes:
Thank you!

PR #2, opened 2016-11-12
Wizardry Refined from Pyramid 3/60, plus some small Magic prereq fixes

dripton:
Wizardry Refined is some tweaks to the spells for Dungeon Fantasy.  I put it in a new file rather than modifying the existing Magic.spl and Dungeon Fantasy.spl, so players have options.

Also found a few little errors in Magic.spl while I was doing this.

### DIFF
--- a/Library/Spells/Magic.spl
+++ b/Library/Spells/Magic.spl
@@ -426,9 +426,6 @@
 				<name compare="is">shape air</name>
 			</spell_prereq>
 			<spell_prereq has="yes">
-				<name compare="is">shape stone</name>
-			</spell_prereq>
-			<spell_prereq has="yes">
 				<name compare="is">shape water</name>
 				<quantity compare="is">1</quantity>
 			</spell_prereq>
@@ -2716,7 +2713,7 @@
 			</prereq_list>
 			<spell_prereq has="yes">
 				<college compare="contains">Body Control</college>
-				<quantity compare="at least">4</quantity>
+				<quantity compare="at least">5</quantity>
 			</spell_prereq>
 			<spell_prereq has="yes">
 				<name compare="is">spasm</name>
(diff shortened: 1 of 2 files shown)


richardwilkes:
There seems to be some confusion in bef2ea6 -- in particular, you state that GCS doesn't count x spells in college prerequisites correctly if the spell with the prerequisite is also in the college. That was the case a release or two ago, but it has since been fixed. I verified this just now using release 4.6. What is interesting is that some of your fixes use the correct number, while others have the number + 1 (as described by your comment). They should all now just be the number as listed in the books. If I'm mistaken on this, please give me an example where it fails to do the correct thing.

dripton:
I tested a bunch of these with 4.5.1.  I will test again with 4.6.

dripton:
Alter Terrain was requiring Shape Stone.  But Magic says it only requires the "all 4 elemental Shape spells".  That would be Earth, Fire, Air, and Water, but not Stone.

I just checked Choke, with gcs 4.6.0.  (With java 1.8.0_101 on 64-bit Ubuntu 16.04, though I doubt it matters with Java.)  Magic says "Magery 1 and at least five Body Control spells, including Spasm."  With the original version of Magic.spl, Choke turned from red (prereqs not met) to black (legal) with only 4 Body Control spells, including Spasm.  With my version, it changed at 5.  So I believe my Choke change is correct.

For Ethereal Body, Magic says "Magery 3 and either Body of Air or six Movement spells."  I read that as (Magery 3) and (Body of Air or 6 Movement spells).  I made it work that way.  Before, it worked like (Magery 3 and Body of Air) or (6 Movement spells.).

For False Memory, Magic says "Forgetfulness and six other Mind Control spells."  With my change (to require Forgetfulness and 7 Mind Control spells", False Memory turns black when I have 7 Mind Control spells including Forgetfulness.  Without my change, it turns black at only 6 Mind Control spells including Forgetfulness.

Great Geas and Lesser Geas and Paralyze Limb are just like Choke and False Memory.

I think these changes are good.  There might be other spells with the same problem that I missed, though.

richardwilkes:
Hmm... maybe I was looking at the wrong thing somehow. I'll check again when I get some time tonight or tomorrow and assuming I was wrong I'll merge the changes. Thanks!

richardwilkes:
OK, reviewing the changes again, they do look good. What threw me off was your comment about needing x+1 to satisfy x, which is not the case. You do need that when certain wording is present (i.e. "Forgetfulness and six *other* Mind Control spells", but not in the general case.

Anyway, thanks again.

PR #5, opened 2016-12-22
New command to increase/decrease skill by levels.

angelforest:
Fixes https://gurpscharactersheet.atlassian.net/projects/GCS/issues/GCS-132
By default, it's set to ctrl+shift+= and ctrl+shift+-, and accessible from "Edit" menu as well.
I'm not sure why git detect "CommandKeyList.md" as entirely changed, I only added two lines for new hot keys.


richardwilkes:
I'm guessing you use a Windows machine and saved the file CommandKeyList.md with Windows line endings rather than unix ones, which is what the GCS source tree uses.

CommandKeyList.md
```diff
@@ -1,61 +1,63 @@
...
-| SHIFT+W | |
-| SHIFT+X | |
-| SHIFT+Y | |
-| SHIFT+Z | |
+
+| Key | Default Mapping |
+| ---: | :--- |
+| - | Decrement (Skill/Spell Points, Advantage Level, Equipment Qty) |
+| = | Increment (Skill/Spell Points, Advantage Level, Equipment Qty) |
+| CTRL+SHIFT+= | Skill Level Increment |
```
richardwilkes: The CTRL or CMD key is implied in this list, as it is platform-specific. Please omit it.

src/com/trollworks/gcs/skill/SkillOutline.java
```diff
@@ -157,6 +158,71 @@ public void increment() {
...
+	}
+
+	@SuppressWarnings("unused")
+	@Override
+	public void incrementSkillLevel() {
+		List<RowUndo> undos = new ArrayList<>();
+		for (Skill skill : new FilteredIterator<>(getModel().getSelectionAsList(), Skill.class)) {
+			if (!skill.canHaveChildren()) {
+				int basePoints = skill.getPoints() + 1;
+				int maxPoints = basePoints + (skill.getDifficulty() == SkillDifficulty.W ? 12 : 4);
```
richardwilkes: What's the purpose of having a maximum points you will add? Given that you just loop until the skill level changes, this seems unnecessary. What am I missing?

src/com/trollworks/gcs/skill/SkillOutline.java
```diff
@@ -157,6 +158,71 @@ public void increment() {
...
+	}
+
+	@SuppressWarnings("unused")
+	@Override
+	public void decrementSkillLevel() {
+		List<RowUndo> undos = new ArrayList<>();
+		for (Skill skill : new FilteredIterator<>(getModel().getSelectionAsList(), Skill.class)) {
+			if (!skill.canHaveChildren()) {
+				int basePoints = Math.max(0, skill.getPoints() - 1);
+				int minPoints = Math.max(0, basePoints - (skill.getDifficulty() == SkillDifficulty.W ? 12 : 4));
```
richardwilkes: The same comment from above seems to apply here, but in reverse. There is the edge case of not being able to go below the point of no points spent, but that's different than what you're doing here.

src/com/trollworks/gcs/skill/SkillOutline.java
angelforest: The idea was to prevent infinite loop in case if level is never increased (for techniques with upper limit). And now I realized I forgot to handle this situation, I should reset points to previous value in that case.

angelforest:
I think I fixed all issues with incorrect level changes, although I didn't have issues with double levels to begin with.  Unfortunately I can't find any differences between my implementation and other commands that use shift key.

richardwilkes:
Looking better. The line endings on the 3 new files (SkillLevelDecrementCommand.java, SkillLevelIncrementCommand.java, and SkillLevelIncrementable.java) are still Windows (CRLF) and should be Unix (LF).

After playing with it a bit, I've determined that the double execution is being caused by something in the OS triggering that specific set of key events twice any time you issue them, which is why selecting it from the menu behaves correctly but using the command key does not. It is likely something specific to macOS. Anyway, I'd suggest changing the command keys to PERIOD for decrement and SLASH for increment, with no SHIFT modifier. This gets around the problem.

Also, please extend this to work with Spells, as they are really just Skills with extra fields and a bit of extra behavior. Thanks!

angelforest:
And done. Thank you for your patience!

richardwilkes:
Thank you! This will be part of the next release.

PR #10, opened 2017-02-28
Make skill containers show total point cost of its children

angelforest:
Few people in discord asked for this.


richardwilkes:
This looks reasonable. I haven't had time to apply the branch and try it out myself -- did you make sure this doesn't break the normal skill point totaling that occurs for the overall sheet?

angelforest:
Yeah, wherever it may happen you already checked if it's container. Otherwise default -1 would screw up total cost anyway.

PR #57, opened 2018-01-28
Add support for Alternate Forms and multiple simultaneous Alternative Abilities

fishface60:
[Sorcerers][sorcerers] can buy up their second most expensive spell to permit simultaneously using two spells. Under the hood this is another Alternative Ability.
This series adds a field to the Advantage Container editor to specify how many should be simultaneously usable and prices them accordingly.

The pricing of Alternate Forms get complicated quickly, especially when it needs to factor in that the shapeshifting might be a racial ability, hence included in itself, and multiple forms of alternate form may exist simultaneously or not applied to a racial template as with [Daniel Dover's Psi-Wars][ddpsiaf], so mechanical assistance is necessary.
This series adds fields to the Advantage Container editor to specify the % discount for the price in excess of the base form, and a text field to name the base form for point calculation purposes.
To use Alternate Forms, create an "Alternate Forms" Advantage container and add "Alternate Form" Advantage Containers as its immediate children, and the "Alternate Forms" container will sum any non-"Alternate Form" Advantage Container items at full price, and only the most expensive "Alternate Form" at full price.
If a base form name is entered in an "Alternate Forms" Advantage Container then its cost will be subtracted from the most expensive "Alternate Form" and apply a % discount if applicable.

Buy levels of the Alternate Form or Morph as applicable as children of the "Alternate Forms" container to price them as per the rules.
I didn't include validation of Alternate Form or Morph being bought since it would limit flexibility and possibly cause rules copyright problems.

I don't program Java much so feedback on programming style is also welcome.

I'm not sure of the etiquette for copyright headers on new files for contributions, so I left it with the boilerplate provided by Eclipse, which asserts the Copyright is yours.

[sorcerers]: http://www.sjgames.com/gurps/books/sorcery/
[ddpsiaf]: http://mailanka.blogspot.co.uk/2016/04/alternate-forms-and-alternative.html

### DIFF
--- a/src/com/trollworks/gcs/advantage/Advantage.java
+++ b/src/com/trollworks/gcs/advantage/Advantage.java
@@ -1054,6 +1054,11 @@ public class Advantage extends ListRow implements HasSourceReference, Switchable
     @Override
     public String getModifierNotes() {
         StringBuilder builder = new StringBuilder();
+        String advantageContainerNotes = mAdvantageContainer.getNotes();
+        if (!advantageContainerNotes.equals("")) { //$NON-NLS-1$
+            builder.append(advantageContainerNotes);
+            builder.append(MODIFIER_SEPARATOR);
+        }
         if (mCR != SelfControlRoll.NONE_REQUIRED) {
             builder.append(mCR);
             if (mCRAdj != SelfControlRollAdjustments.NONE) {
--- a/src/com/trollworks/gcs/advantage/Advantage.java
+++ b/src/com/trollworks/gcs/advantage/Advantage.java
@@ -27,7 +27,6 @@ import com.trollworks.gcs.widgets.outline.ListRow;
 import com.trollworks.gcs.widgets.outline.RowEditor;
 import com.trollworks.gcs.widgets.outline.Switchable;
 import com.trollworks.toolkit.annotation.Localize;
-import com.trollworks.toolkit.collections.FilteredIterator;
 import com.trollworks.toolkit.io.xml.XMLReader;
 import com.trollworks.toolkit.io.xml.XMLWriter;
 import com.trollworks.toolkit.ui.image.StdImage;
@@ -132,9 +131,10 @@ public class Advantage extends ListRow implements HasSourceReference, Switchable
     private String                     mReference;
     private String                     mOldPointsString;
     private AdvantageContainerType     mContainerType;
+    private AdvantageContainer         mAdvantageContainer;
     private ArrayList<WeaponStats>     mWeapons;
     private ArrayList<Modifier>        mModifiers;
-    private boolean                    mRoundCostDown;
+    boolean                            mRoundCostDown;
     private boolean                    mDisabled;
 
     /**
@@ -152,6 +152,7 @@ public class Advantage extends ListRow implements HasSourceReference, Switchable
         mLevels = -1;
         mReference = ""; //$NON-NLS-1$
         mContainerType = AdvantageContainerType.GROUP;
+        mAdvantageContainer = SummativeAdvantageContainer.getInstance();
         mWeapons = new ArrayList<>();
         mModifiers = new ArrayList<>();
     }
@@ -178,6 +179,7 @@ public class Advantage extends ListRow implements HasSourceReference, Switchable
         mDisabled = advantage.mDisabled;
         mReference = advantage.mReference;
         mContainerType = advantage.mContainerType;
+        mAdvantageContainer = advantage.mAdvantageContainer;
         mWeapons = new ArrayList<>(advantage.mWeapons.size());
         for (WeaponStats weapon : advantage.mWeapons) {
             if (weapon instanceof MeleeWeaponStats) {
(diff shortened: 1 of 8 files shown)


richardwilkes:
- When Alternative Abilities is chosen, a note should be visible on the container stating how many simultaneous abilities can be active at once
- For Alternative Forms, I'm not following where you're getting the rules for this from. The primary pricing information I see is this, from the Alternative Form advantage on B83-84:

> If you have a single Alternate Form, it costs 15 points for a racial template worth no more than your native racial template. A more powerful form costs 15 points plus 90% of the difference in cost between your native template and that of your Alternate Form.
>
> If you have multiple forms, pay full cost for the most expensive form. The less powerful Alternate Forms cost a flat 15 points apiece. Minimum cost per form is still 15 points.
> 

This code does not price the form this way at all. Could you give me some more background on where you're getting the pricing and how it reconciles with the Basic book?

fishface60:
> * When Alternative Abilities is chosen, a note should be visible on the container stating how many simultaneous abilities can be active at once

I'll add that in the next version of this pull request.

> * For Alternative Forms, I'm not following where you're getting the rules for this from. The primary pricing information I see is this, from the Alternative Form advantage on B83-84:
> > If you have a single Alternate Form, it costs 15 points for a racial template worth no more than your native racial template. A more powerful form costs 15 points plus 90% of the difference in cost between your native template and that of your Alternate Form.
> > 
> > If you have multiple forms, pay full cost for the most expensive form. The less powerful Alternate Forms cost a flat 15 points apiece. Minimum cost per form is still 15 points.

I'm mostly getting my rules from there, with the small exception of not hard-coding the -10% to the difference in price to the native template with just one alternate form, since as described in Powers p75 if you add "Once On, Stays On" the discount doesn't apply.

If I've misinterpreted things I'd appreciate it if you point out where my understanding is flawed.

The modifiers on the cost of Alternate Form and Morph are mostly modifiers to the cost of the 15 per form or 100 for Morph, so it may be the most appropriately modelled as a separate Advantage to the template you want to shape-shift into.
You're effectively buying a number of Shape-shifting slots. 15 points each for Alternate Form and 100 for equal to IQ with Morph.

Having bought the slots it's the cost of the forms that is left. For Morph, multiple Alternate Forms, or a single Alternate form with "Once On, Stays On" this can be handled with a group that costs as much as the most expensive member if you put your native template in it.

1.  If your Alternate Forms cost less or equal to your native template the most expensive item is the native template so the whole group would cost as much as the native template which meets the requirement of the extra cost of alternate forms that cost less than the native template is zero.
2.  If any of your Alternate Forms cost more than your native template the group costs the most expensive Alternate Form meeting the description of that second paragraph. It's a little ambiguous whether the "Full Price" includes the 15 points but the intent seems to be each extra form costs 15.

This doesn't work out so well for "One alternate form without Once On, Stays On" case, since it needs to know which is the base form and which is the Alternate form, and it can't just assume the cheapest is the base form since you might have an alternate form that's cheaper than your native template so should cost zero.

This is handled by having the text field to name the base form, and charging members of the Alternate Forms group which aren't Alternate Forms full price (so the base form isn't priced as an alternate form) and subtracting the price of the base form of the cost  of Alternate Forms before applying the -10% discount.

This then gets complicated by the fact that you can have a native template that includes shape shifting:

> Shapeshifting Races: When creating an entire race that has Alternate Form, work out the details of the Alternate Form last. Total the cost of all the race’s traits other than Alternate Form, subtract this total from the cost of the template the race transforms into, and use the difference to calculate the cost of Alternate Form for the race. Add the cost of Alternate Form to that of the race’s other abilities to  determine final racial cost.

If you've only got one form of Shape-shifting then you can add your template without the shape-shifting as the base form in the Alternate Forms group, but if you were a lycanthropic elementalist you'd want to keep the two separate and your base form can't exist in both.
To handle this the Alternate Forms group creates a copy of the Base Form without the shape-shifting and subtracts the cost of that from the cost of Alternate Forms.

richardwilkes:
OK, I think I understand the costs now, although I don't believe I saw correct pricing when I was playing with it yesterday. I'll have to try again now that you've explained what is intended.

I have not yet had time to really sit down and go over the changes you've made to the code, although one thing leaped out at me immediately that I'm not thrilled with, and that was the addition of another "advantage container" object. Can you not roll the behavior into the AdvantageContainerType combined with looking at any modifiers attached to it for the special cases (which all seem to be modifiers, aren't they?)? Like I said, I've only briefly glanced at the code so far, so I might be off-base on this, but I would prefer to keep such changes to a minimum, especially since I am currently working on a re-write of GCS to fix some fundamental problems, such as the current inability to automatically update existing sheets with library data changes.

fishface60:
> OK, I think I understand the costs now, although I don't believe I saw correct pricing when I was playing with it yesterday. I'll have to try again now that you've explained what is intended.

I'll have a go at pricing up some of the examples in the book to verify it produces the same results. They were correct to my understanding earlier in development but it's possible bugs crept in after I moved my focus to the user interface or my understanding of the pricing changed.

> I have not yet had time to really sit down and go over the changes you've made to the code, although one thing leaped out at me immediately that I'm not thrilled with, and that was the addition of another "advantage container" object. Can you not roll the behavior into the AdvantageContainerType combined with looking at any modifiers attached to it for the special cases (which all seem to be modifiers, aren't they?)?

I tried adding a modifier in an earlier version, but abandoned it when I encountered difficulties of not having enough context available to say that the modifier is only available when in an Advantage Container, and insufficient familiarity and confidence with the code base and Java to work out a good way to pass it through.

I'll see if I can work out a way to model it appropriately with generic modifiers, or a better way of having Container Type specific modifiers.

> Like I said, I've only briefly glanced at the code so far, so I might be off-base on this, but I would prefer to keep such changes to a minimum, especially since I am currently working on a re-write of GCS to fix some fundamental problems, such as the current inability to automatically update existing sheets with library data changes.

I'll see what I can do. I'm not sure how much time I'll have available to work on this.

richardwilkes:
I'm going to assume you've decided not to proceed with this, given the nearly 2 years that have elapsed. If that's not the case, feel free to reopen.

PR #103, opened 2019-06-05
Support for Fantasy Grounds text/xml export

crnormand:


src/com/trollworks/gcs/advantage/Advantage.java
```diff
@@ -48,7 +48,7 @@
 public class Advantage extends ListRow implements HasSourceReference, Switchable {
     @Localize("Advantage")
     @Localize(locale = "de", value = "Vorteil")
-    @Localize(locale = "ru", value = "Преимущество")
+    @Localize(locale = "ru", value = "ÐŸÑ€ÐµÐ¸Ð¼ÑƒÑ‰ÐµÑ�Ñ‚Ð²Ð¾")
```
richardwilkes: This looks like your editor mangled the UTF-8 encoding

src/com/trollworks/gcs/advantage/Advantage.java
```diff
@@ -1040,6 +1040,21 @@ public boolean setModifiers(List<Modifier> modifiers) {
         return false;
     }
 
+    /**
+     * Find the first modifier that matches 'name'.  
+     *
+     * @param name case insensitive
+     * @return
```
richardwilkes: @return by itself isn't useful. Please change this to match other uses... something like:
```
/**
 * @param name The name to match against. Case-insensitive.
 * @return The first modifier that matches the name.
 */

File: src/com/trollworks/gcs/character/TextTemplate.java [hunk 3]
```diff
@@ -574,19 +644,37 @@ private void emitKey(BufferedReader in, BufferedWriter out, String key, File bas
...
             } else {
                 writeEncodedText(out, String.format(UNIDENTIFIED_KEY, key));
             }
             break;
         }
     }
 
+    private void setOnlyCategories(String key) {
+        String[] categories = key.substring(KEY_ONLY_CATEGORIES.length()).split(UNDERSCORE);
+        mOnlyCategories.addAll(Arrays.asList(categories));
```
richardwilkes: Perhaps `mOnlyCategories` and `mExcludedCategories` (below) should be emptied before setting the values here? Otherwise, you can only ever add to the lists.

I see later on that you actually implicitly clear the values at certain points... I guess that's reasonable, too.

src/com/trollworks/gcs/character/TextTemplate.java
```diff
@@ -742,13 +858,43 @@ private void processHitLocationLoop(BufferedWriter out, String contents) throws
...
+        for (Equipment equipment : mSheet.getCharacter().getEquipmentIterator()) {
+            if (equipment.isEquipped()) {
+                for (Feature feature : equipment.getFeatures()) {
+                    if (feature instanceof DRBonus) {
+                        String locationKey = entry.getLocation().getKey();
+                        if (locationKey.equals(HitLocation.VITALS.getKey())) {
+                            // Assume Vitals uses the same equipment as Torso
+                            locationKey = HitLocation.TORSO.getKey();
+                        }
+                        if (locationKey.endsWith(((DRBonus) feature).getLocation().name())) {
```
richardwilkes: You can and should be doing direct `==` comparison with the HitLocation -- they are all static final values.

src/com/trollworks/gcs/character/TextTemplate.java
```diff
@@ -1014,48 +1181,333 @@ private void processMeleeLoop(BufferedWriter out, String contents) throws IOExce
...
+     * and then possibly one time for each different "attack mode" that the weapon can support.
+     * e.g. Weapon Name: Spear, attack modes "1 Handed" and "2 Handed"
+     */
+    private void processHierarchicalMeleeLoop(BufferedWriter out, String contents) throws IOException {
+        int                                          length           = contents.length();
+        StringBuilder                                keyBuffer        = new StringBuilder();
+        boolean                                      lookForKeyMarker = true;
+        int                                          counter          = 0;
+        HashMap<String, ArrayList<MeleeWeaponStats>> weaponsMap       = new HashMap<String, ArrayList<MeleeWeaponStats>>();
+        HashMap<String, MeleeWeaponStats>            weapons          = new HashMap<String, MeleeWeaponStats>();
```
richardwilkes: Replace the right side of these two declarations with just `new HashMap<>()`

src/com/trollworks/gcs/character/TextTemplate.java
```diff
@@ -1092,63 +1544,39 @@ private void processRangedLoop(BufferedWriter out, String contents) throws IOExc
...
+        int                  length           = contents.length();
+        StringBuilder        keyBuffer        = new StringBuilder();
+        boolean              lookForKeyMarker = true;
+        int                  counter          = 0;
+        // Create child-to-parent maps to determine where items are being stored.
+        // Used by KEY_LOCATION
+        ArrayList<List<Row>> children         = new ArrayList<List<Row>>();
+        ArrayList<Equipment> parents          = new ArrayList<Equipment>();
+
+        ArrayList<Equipment> equipmentList    = new ArrayList<Equipment>();
```
richardwilkes: The above 3 lines should have the redundant type specification eliminated

src/com/trollworks/gcs/advantage/Advantage.java
crnormand: Agreed.   I think I changed the encoding AFTER I had downloaded the code.   Sorry.   How can I revert this?

src/com/trollworks/gcs/character/TextTemplate.java
crnormand: Yes, there were a few ways I could do it, so I chose to allow the template to add to either list using @ONLY_CATEGORIES_ and @EXCLUDE_CATEGORIES_ prior to executing any loop.   Either on one line, or multiple.

src/com/trollworks/gcs/advantage/Advantage.java
richardwilkes: As long as your editor/IDE is set to the right mode, you should be able to just copy and paste the deleted line above and then re-push it.

src/com/trollworks/gcs/character/TextTemplate.java
crnormand: I cannot.   The HitLocation from the equipment is com.trollworks.gcs.character.HitLocation and the HitLocation from the DRBonus is com.trollworks.gcs.feature.HitLocation.   They are not the same object.    

The key for com.trollworks.gcs.character.HitLocation is created from strings from com.trollworks.gcs.character.Armor, which were created by prepending DR_PREFIX ("gcs.dr.") to the  com.trollworks.gcs.feature.HitLocation enum.    Which is why I had to go through such lengths to get them to match.   However, since the character.HitLocaiton name is built from the feature.HitLocation class, the string check will always be correct.

richardwilkes: Ah. I had forgotten about that. OK, ignore me! :-)

src/com/trollworks/gcs/advantage/Advantage.java
crnormand: I believe I fixed this with the latest commit.

Issue #81, opened 2018-09-18
Features Category bonus


Zuljita:
It'd be handy to be able to set a talent to give +1 to things like "Skills whose category is".  Particularly useful for things like spells or imbuements.

crnormand:
Similar to #35.    Pull Requests #117 (Skills/Techniques), #118 (Spells) and #119 (Weapons Bonuses) now allow you to set a criteria based on categories.   When you are building the Talent, you can set the particular bonus based on the category.

Issue #35, opened 2018-01-15
Allow a modifier to affect skills with a given category tag


richardwilkes:
Some of the advantages apply modifiers to a broad category, e.g. a bonus to all thieving skills, or bonus weapon damage to all melee attacks. For these, it would be easiest to configure if advantages had "Gives a Skill Level Bonus to Skills Whose Category is X" and "Gives a Weapon Damage Bonus" to Weapons Whose Required Skill has Category X".

crnormand:
Pull Requests https://github.com/richardwilkes/gcs/pull/117 (Skills/Techniques), https://github.com/richardwilkes/gcs/pull/118 (Spells) and https://github.com/richardwilkes/gcs/pull/119 (Weapons Bonuses) now allow you to set a criteria for searching on categories.

Issue #46, opened 2018-01-15
Support for more specific bonuses


richardwilkes:
There are cases where a bonus should only be applied to a particular weapon. Right now, bonuses get applied to all weapons of a given type.

crnormand:
Although not perfect, this can be accomplished with https://github.com/richardwilkes/gcs/pull/119.   Add a new category to the weapon (e.g. "Bonded"), and then build the bonus so that it only applies to weapons with that category (e.g. "+1 Damage to all weapons whose category is "Bonded").

PR #104, opened 2019-06-07
Added Category matching for Skill level and Weapon Damage bonuses

crnormand:
Looking at https://github.com/richardwilkes/gcs/issues/81, I thought I could add a criteria to match against the categories.   I was able to add it for Skill bonuses and Weapon Bonuses, but was uncertain how to handle Spell bonuses.   I am uncertain what 'mMatchType' is, and how adding a category criteria would affect getKey() and the XML read/write methods.   So i left that alone for the moment.

But it works for Skills (and techniques) and Weapons bonuses.  You can make "quick" Talents for skills by selecting "whos name is anything" and "category is" and then entering a category (e.g. "Animal") and all skills with that category will gain the bonus.

You can also make a quick "Striking ST" kind of bonus by creating a WeaponBonus and set "whos required skill name is anything" and "category is" "Melee weapon".   All melee weapons will now get the bonus damage.

Unfortunately, the OP requested the ability specifically for Spells... so I will continue to investigate.

Oh, and this seems to be directly on point for https://github.com/richardwilkes/gcs/issues/35

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -37,6 +37,10 @@
 			<attribute name="module" value="true"/>
 		</attributes>
 	</classpathentry>
-	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11"/>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER">
+		<attributes>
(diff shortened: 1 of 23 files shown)


crnormand:
This might also work for https://github.com/richardwilkes/gcs/issues/46.   You would have to add a particular category to the weapon, and then use this feature to only apply the weapon bonus to that category.

PR #109, opened 2019-07-01
 Support for Row based tooltip in Mulitcell 

crnormand:
#3 (in a series)

Ok, well, I created a branch and pushed it, but I didn't do it soon enough.   I guess I should have created a branch for my previous pull request as well.

Sorry... still learning the 'git' process.

However, I must admit that many of my changes will be sequential, meaning that you will need to accept them in order, since they will build on themselves.

So I can wait until you accept these pull requests, and then create a new one with the next set of changes.

### DIFF
--- a/src/com/trollworks/gcs/widgets/outline/MultiCell.java
+++ b/src/com/trollworks/gcs/widgets/outline/MultiCell.java
@@ -205,7 +205,6 @@ public class MultiCell implements Cell {
         return Cursor.getDefaultCursor();
     }
 
-    /** Allow a satified row to display a tooltip */
     @Override
     public String getToolTipText(Outline outline, MouseEvent event, Rectangle bounds, Row row, Column column) {
         ListRow theRow = (ListRow) row;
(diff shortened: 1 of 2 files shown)


src/com/trollworks/gcs/widgets/outline/MultiCell.java
```diff
@@ -208,7 +208,18 @@ public Cursor getCursor(MouseEvent event, Rectangle bounds, Row row, Column colu
     @Override
     public String getToolTipText(Outline outline, MouseEvent event, Rectangle bounds, Row row, Column column) {
         ListRow theRow = (ListRow) row;
-        return theRow.isSatisfied() ? null : theRow.getReasonForUnsatisfied();
```
richardwilkes: Same comment as in the other PR...

crnormand:
Both are relevant.   I am trying to stage my PRs in order, so you can inspect and accept smaller, more manageable chunks.   Unfortunately, I was not good at branching, and my commits appeared in multiple places.   I put the same code change in both PRs, so you should be able to merge them with no issues.

PR #113, opened 2019-07-12
Added tooltips for Spells & Weapon Damage

crnormand:
This adds tooltips for Weapon Damage and Spells.   It is a slightly bigger PR than before, but it follows the same pattern as Skill.

1.  I had to increase the tooltip time so that the user has time to actually read the bonuses.   I wasn't certain where you would want that kind of change, so I put it in GCS.java.

2.  The other GURPSCharacter "get*Bonus*" methods had to be updated to include the tooltip.

3.  In the previous PR, I did not completely update the PrerequisitesThread class.   I missed a few places where the parent needed to be set.

4.  Similar to the Skill PR, Spell's now use an instance of SkillLevel to maintain their level and relative level.   Also, the Spell calculation methods needed to be updated with tooltips.

5.  SpellColumn now shows tooltips for the Level and Relative Level columns.

6.  WeaponColumn now shows tooltips for the Damage column.

7.  WeaponDisplayRow and WeaponStats have been updated to calculate a tooltip while determining damage.

### DIFF
--- a/src/com/trollworks/gcs/app/GCS.java
+++ b/src/com/trollworks/gcs/app/GCS.java
@@ -48,6 +48,8 @@ import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.StringTokenizer;
 
+import javax.swing.ToolTipManager;
+
 /** The main entry point for the character sheet. */
 public class GCS {
     @Localize("GURPS Character Sheet")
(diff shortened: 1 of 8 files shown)


src/com/trollworks/gcs/app/GCS.java
```diff
@@ -233,6 +235,9 @@ private static void initialize() {
         Fonts.loadFromPreferences();
         App.setAboutPanel(AboutPanel.class);
         registerFileTypes(new GCSFileProxyCreator());
+
+        // Increase TooTip time so the user has time to read the skill modifiers
+        ToolTipManager.sharedInstance().setDismissDelay(60000);
```
richardwilkes: This is fine for the moment, but it really should be a preference the user can change

PR #114, opened 2019-07-12
Prefs

crnormand:
Great suggestion.   I had to get up early to drive my wife to the airport, so I am wired on caffeine and decided to hammer this out before the coffee wore off.

I created a new preferences tab, "System".   It only has the one item on it (so it centers it on the screen), but it is fully functional.

image


crnormand:
You mean like this:

crnormand:
After looking through other code, I assume that is what you meant, so that is what I did.  ;-)

PR #117, opened 2019-07-13
Added bonus/feature by category for Skills

crnormand:
Added bonuses by category for Skills.   This one is kind of messy because of the necessity of passing the categories around.

1.  The "get*Bonus*" methods in GURPSCharacter now accept a category string (actually, a single string of all categories separated by commas), and when they are comparing the bonus to see if it matches, they now also check the category criteria.

2.  To shorten some of the criteria checking code, they can now respond directly if they are "an exact match" kind of criteria (Is), or an "anything goes" kind (isAnything).

3.  SkillBonus gained a category criteria and all of the internal methods were updated to include it (object creation, copy, XML read/write, etc.).   Also, a SkillBonus can now determine if it matchesCategories().

4.  The SkillBonusEditor gained a category criteria row.

5.  Skill was updated to include its category in all of its level calculation methods.

6.  And the SkillEditor was modified to pass in the category when determining level.

7.  And finally, Technique and the TechniqueEditor were modified to include the categories as well, since they subclass from Skill.

7a.   And I forgot a tooltip method on the TechniqueEditor, so it will show tooltips just like the SkillEditor.

I am sorry this one is hairier than the previous PRs.   I tried to whittle out as much as possible, but except for #2 and #7a (above), everything is necessary to make this work.

And unfortunately, the following PRs (for Spells and WeaponDamage) are going to be about the same.

### DIFF
--- a/src/com/trollworks/gcs/character/GURPSCharacter.java
+++ b/src/com/trollworks/gcs/character/GURPSCharacter.java
@@ -66,6 +66,7 @@ import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Set;
 
 /** A GURPS character. */
 public class GURPSCharacter extends DataFile {
@@ -2722,10 +2723,10 @@ public class GURPSCharacter extends DataFile {
      * @param id                      The feature ID to search for.
      * @param nameQualifier           The name qualifier.
      * @param specializationQualifier The specialization qualifier.
-     * @parem categoryQualifier The categories qualifier
+     * @param categoryQualifier       The categories qualifier
      * @return The bonus.
      */
-    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier) {
+    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, Set<String> categoryQualifier) {
         return getSkillComparedIntegerBonusFor(id, nameQualifier, specializationQualifier, categoryQualifier, null);
     }
 
@@ -2733,11 +2734,11 @@ public class GURPSCharacter extends DataFile {
      * @param id                      The feature ID to search for.
      * @param nameQualifier           The name qualifier.
      * @param specializationQualifier The specialization qualifier.
-     * @parem categoryQualifier The categories qualifier
-     * @param toolTip The toolTip being built
+     * @param categoryQualifier       The categories qualifier
+     * @param toolTip                 The toolTip being built
      * @return The bonus.
      */
-    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier, StringBuilder toolTip) {
+    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, Set<String> categoryQualifier, StringBuilder toolTip) {
         int                total = 0;
         ArrayList<Feature> list  = mFeatureMap.get(id.toLowerCase());
         if (list != null) {
--- a/src/com/trollworks/gcs/character/GURPSCharacter.java
+++ b/src/com/trollworks/gcs/character/GURPSCharacter.java
@@ -2722,17 +2722,19 @@ public class GURPSCharacter extends DataFile {
      * @param id                      The feature ID to search for.
      * @param nameQualifier           The name qualifier.
      * @param specializationQualifier The specialization qualifier.
+     * @parem categoryQualifier The categories qualifier
      * @return The bonus.
      */
     public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier) {
-        return getSkillComparedIntegerBonusFor(id, nameQualifier, specializationQualifier, null);
+        return getSkillComparedIntegerBonusFor(id, nameQualifier, specializationQualifier, categoryQualifier, null);
     }
 
     /**
      * @param id                      The feature ID to search for.
      * @param nameQualifier           The name qualifier.
      * @param specializationQualifier The specialization qualifier.
-     * @param toolTip                 The toolTip being built
+     * @parem categoryQualifier The categories qualifier
+     * @param toolTip The toolTip being built
      * @return The bonus.
      */
     public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier, StringBuilder toolTip) {
--- a/src/com/trollworks/gcs/character/GURPSCharacter.java
+++ b/src/com/trollworks/gcs/character/GURPSCharacter.java
@@ -2724,7 +2724,7 @@ public class GURPSCharacter extends DataFile {
      * @param specializationQualifier The specialization qualifier.
      * @return The bonus.
      */
-    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier) {
+    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier) {
         return getSkillComparedIntegerBonusFor(id, nameQualifier, specializationQualifier, null);
     }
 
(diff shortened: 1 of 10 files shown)


src/com/trollworks/gcs/character/GURPSCharacter.java
```diff
@@ -2724,7 +2724,7 @@ public int getIntegerBonusFor(String id, StringBuilder toolTip) {
      * @param specializationQualifier The specialization qualifier.
      * @return The bonus.
      */
-    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier) {
+    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier) {
         return getSkillComparedIntegerBonusFor(id, nameQualifier, specializationQualifier, null);
```
richardwilkes: This just calls itself, in an infinite loop

src/com/trollworks/gcs/character/GURPSCharacter.java
```diff
@@ -2724,7 +2724,7 @@ public int getIntegerBonusFor(String id, StringBuilder toolTip) {
      * @param specializationQualifier The specialization qualifier.
      * @return The bonus.
      */
-    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier) {
+    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier) {
```
richardwilkes: Need to update the JavaDoc to include the new parameter

src/com/trollworks/gcs/character/GURPSCharacter.java
```diff
@@ -2735,14 +2735,14 @@ public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, Stri
      * @param toolTip                 The toolTip being built
      * @return The bonus.
      */
-    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, StringBuilder toolTip) {
+    public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, String specializationQualifier, String categoryQualifier, StringBuilder toolTip) {
```
richardwilkes: Javadoc again

src/com/trollworks/gcs/feature/SkillBonus.java
```diff
@@ -98,28 +103,41 @@ public String getKey() {
         return buffer.toString();
     }
 
+    public boolean matchesCategories(String categories) {
+        String[] cats = categories.split(COMMA);
+        for (String category : cats) {
+            if (mCategoryCriteria.matches(category.trim())) {
```
richardwilkes: Should probably do a case-insensitive comparison

src/com/trollworks/gcs/skill/Skill.java
```diff
@@ -869,7 +869,7 @@ private SkillDefault getBestDefault(SkillDefault excludedDefault) {
                         int level = skillDefault.getType().getSkillLevel(character, skillDefault, excludes);
                         if (skillDefault.getType().isSkillBased()) {
                             String name = skillDefault.getName();
-                            level -= character.getSkillComparedIntegerBonusFor(ID_NAME + ASTERISK, name, skillDefault.getSpecialization());
+                            level -= character.getSkillComparedIntegerBonusFor(ID_NAME + ASTERISK, name, skillDefault.getSpecialization(), "");  //$NON-NLS-1$
```
richardwilkes: Doesn't this need to know the categories to do the right thing?

src/com/trollworks/gcs/feature/SkillBonus.java
crnormand: Actually, it is already case insensitive.   The StringCriteria code handles that.

src/com/trollworks/gcs/character/GURPSCharacter.java
crnormand: Duh.... sorry, was trying to cull out just these changed from my other instance of Eclipse.   It has been fixed in the latest commit.

src/com/trollworks/gcs/skill/Skill.java
crnormand: I was uncertain.    The problem is that SkillDefault did not have a category, so I assumed none.   But I see your point, on line 872, we are trying to find the level of the skill that we are basing this default on... which is different that the Skill from which we are trying to calculate the SkillDefault for.

I think I have it fixed.   It looks up the Skill for the SkillDefault and uses that Skill's categories when asking the character for the correct level for that skill..

Unfortunately, it doesn't correctly update the tooltip :-(    The default is calculated before the Skill and any modifiers that have affected the default are lost.   Not a huge problem (i can already see how to fix it), but I will leave that to the next PR.

crnormand:
I kept the categories as a string because of the editors.  While in the editors, the category is just a string, so I would have to encode it into the set of categories.   To me it seemed 6 of one, 1/2 dozen of the other.

I do agree that if we keep the "categories as string" parameter, that I should consolidate all of the category composition/decomposition/matching/etc. login in the same class.        Which is what I have done with this latest commit.

I would like to keep it this way (as a String vs. Set<String>) with your permission.   All of the code for composing and decomposing the categories is kept in ListRow, now.  If you feel strongly against, I can refactor and have the editors call the composition method to make the Set<String> instead.

Concerning the two category matching methods, they actually support slightly different functionality.   hasCategory() is solely for the TextTemplate output (which has the funky logic of being able to match on the category name, or the name up to ":").   The other (now implemented in ListRow.matchesCategories()) is used by the criteria system, which could include any kind of "is/is not/begins with", etc.    

I could change the hasCategory() method to accept a StringCriteria insead, but I wasn't sure we would want to include the funky logic check with our generic criteria matching.   It was specifically designed for a FantasyGrounds feature where are can split advantages and equipment by category.   When we want to split "coins" from the rest of the inventory, there was a problem because many of the coins in the data libraries have the category "Money: GURPS Standard Medieval Coins", but the DFRPG library has the category set to "Money".   So being able to match on "Money" was the only way, with the data as is, catch as many of the coin instances as possible.

Hence why there are 2 different "matching" methods.

src/com/trollworks/gcs/character/GURPSCharacter.java
```diff
@@ -2722,17 +2722,19 @@ public int getIntegerBonusFor(String id, StringBuilder toolTip) {
      * @param id                      The feature ID to search for.
      * @param nameQualifier           The name qualifier.
      * @param specializationQualifier The specialization qualifier.
+     * @parem categoryQualifier The categories qualifier
```
richardwilkes: `@parem` -> `@param` ... surprised Eclipse didn't complain about this

src/com/trollworks/gcs/character/GURPSCharacter.java
```diff
@@ -2722,17 +2722,19 @@ public int getIntegerBonusFor(String id, StringBuilder toolTip) {
...
-        return getSkillComparedIntegerBonusFor(id, nameQualifier, specializationQualifier, null);
+        return getSkillComparedIntegerBonusFor(id, nameQualifier, specializationQualifier, categoryQualifier, null);
     }
 
     /**
      * @param id                      The feature ID to search for.
      * @param nameQualifier           The name qualifier.
      * @param specializationQualifier The specialization qualifier.
-     * @param toolTip                 The toolTip being built
+     * @parem categoryQualifier The categories qualifier
```
richardwilkes: Same here

crnormand:
Understood, and I will get on it this evening.

src/com/trollworks/gcs/character/GURPSCharacter.java
crnormand: Ok, I think I have spell checking enabled correctly... although that is no guarantee that I will spell things right.   ;-)

richardwilkes: The JavaDoc verifier should actually be throwing a fit on these... but I must admit I haven't specifically tested that theory out.

crnormand:
Done.   The latest commit passes the categories as Set<String> instead of string.

PR #118, opened 2019-07-21
Added Category filtering to Spell Bonuses

crnormand:
This builds off of (and partially copies) the Skill categories code.   The biggest issue here was the multiple ways you could add a spell bonus (by college, by name, by Power, etc.).

I built a large testcase that created bonuses for all of the combinations of college/power/name and category, to verify that the code works (and doesn't interfere with previous bonuses).  So on this, you will have to trust me ;-)  

image

The final PR will contain the code to support weapon bonuses by category.

### DIFF
--- a/src/com/trollworks/gcs/character/GURPSCharacter.java
+++ b/src/com/trollworks/gcs/character/GURPSCharacter.java
@@ -2756,8 +2756,9 @@ public class GURPSCharacter extends DataFile {
     }
 
     /**
-     * @param id        The feature ID to search for.
-     * @param qualifier The qualifier. #param categories The categories qualifier
+     * @param id         The feature ID to search for.
+     * @param qualifier  The qualifier.
+     * @param categories The categories qualifier
      * @return The bonus.
      */
     public int getSpellComparedIntegerBonusFor(String id, String qualifier, Set<String> categories, StringBuilder toolTip) {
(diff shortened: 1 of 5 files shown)


src/com/trollworks/gcs/character/GURPSCharacter.java
```diff
@@ -2757,17 +2757,17 @@ public int getSkillComparedIntegerBonusFor(String id, String nameQualifier, Stri
 
     /**
      * @param id        The feature ID to search for.
-     * @param qualifier The qualifier.
+     * @param qualifier The qualifier. #param categories The categories qualifier
```
richardwilkes: The parameter should be on its own line and start with @, not #

src/com/trollworks/gcs/feature/SpellBonus.java
```diff
@@ -21,34 +21,40 @@
...
 
     /** Creates a new spell bonus. */
     public SpellBonus() {
         super(1);
-        mAllColleges  = true;
-        mMatchType    = TAG_COLLEGE_NAME;
-        mNameCriteria = new StringCriteria(StringCompareType.IS, ""); //$NON-NLS-1$
+        mAllColleges      = true;
+        mMatchType        = TAG_COLLEGE_NAME;
+        mNameCriteria     = new StringCriteria(StringCompareType.IS, ""); //$NON-NLS-1$
```
richardwilkes: Probably should use EMPTY here, too

Issue #120, opened 2019-07-21
Allow "User Descriptions" for Advantages on character sheet


crnormand:
When using any of the "Powers as Magic" systems (Powers - Divine Favor, Thaumatology - Sorcery, etc.), your character will start to collect a lot of "powers" (Advantages with modifiers).

In most cases, you want to be able to describe how the spell works, and what it does.  For example: 

"Confundus Charm; causes the victim to become confused and disoriented.  Contest your Will (-1 per hex) vs their HT or become “Dazed” for margin of failure minutes."

Which you can add to the Notes field of the Advantage, and it will display with the Advantage (and any Melee/Ranged attacks that you define with it).

You can see the Advantage "Confundo" highlighted in this picture, and you can see that the Confundo Ranged Attack (which includes the Notes) on the left side:
![image](https://user-images.githubusercontent.com/8931036/61596368-7e2df880-abd0-11e9-8ee7-5c794c46bfbd.png)

But the real problem comes when the spell description isn't a simple 1-liner, but instead something like this:

"The druid can direct existing vegetation, roots, etc. to entangle and restrain everyone within an area.  Roll against Innate Attack (Gaze) to target a hex. If you miss, see Scatter (p. B414). The binding lasts for 10 seconds, continuing to entangle new people who blunder into it.  Victims inside the area may try to Dive (+3 Dodge, then Prone) or Retreat (+3 Dodge, 1 step back), to get out of the area of effect.  Anyone still inside the area is grappled (see p. 370) and rooted in place.  They cannot select the Move or Change Posture maneuvers or change facing, and is at -4 to DX.   To break free, the victim must win a Quick Contest of ST or Escape skill against the ST of your Binding. Each attempt takes one second. If the victim fails to break free, he loses 1 FP but may try again.   Those that do not break free grow more tightly entangled, with +1 to the ST of their bonds for each additional second they remain grappled."

Trying to add this to the Notes field results in this:
![image](https://user-images.githubusercontent.com/8931036/61596384-da911800-abd0-11e9-9b6c-49d1791506bd.png)

It would be nice if we could a define a separate text field to contain "larger than 1 line" text.    And optionally display it in a tooltip using the new Tooltip feature (https://github.com/richardwilkes/gcs/pull/112, https://github.com/richardwilkes/gcs/pull/113 & https://github.com/richardwilkes/gcs/pull/115), and allow it to be exported as part of the text output feature.

richardwilkes:
[...]

PR #123, opened 2019-07-25
Added a User Description to Advantage

crnormand:
Ok, I really need your guidance on this one.   I added a "user description" to Advantage (as discussed in https://github.com/richardwilkes/gcs/issues/120).  

I wanted to use the Notes editor (and therefore a Note), but I wasn't too keen on adding a Note object to every Advantage.  And besides, Notes seem to need to be in an outline (collection) of Notes, so I was going to have to do some hacking to add a single Note.

However, to maintain the same "look", I wanted to reuse as much of the NoteEditor as I could.  Alas, I could not.   I made a temporary Note to try to fool the editor, but it is all closely tied to having a datafile and an owner (outline),   As some point (after faking out as much as I could), it just became too much, so I just copied the GUI code into the openUserDescDialog() method.

Yes, the OO programmer inside my head was screaming at me... but I just wanted to get it done.  It bothered me so much (as I was typing up this PR), that I refactored as much as I could into a static method.

I didn't create localized versions of the button label "User Desc.", since I wasn't certain how to handle the abbreviation.   Of course, the dialog title has been localized.

Where should I go from here?   This code works (now, after the second commit, I had to fix the "dirty" flag for the user description), and externally, looks good.

Once you are ok with this (however long that takes), I would like to add it to the tooltip code and add a property to turn it on/off, and add it to the text output code, so I can export this information to Fantasy Grounds.

### DIFF
--- a/src/com/trollworks/gcs/advantage/Advantage.java
+++ b/src/com/trollworks/gcs/advantage/Advantage.java
@@ -112,6 +112,8 @@ public class Advantage extends ListRow implements HasSourceReference, Switchable
     public static final String         ID_WEAPON_STATUS_CHANGED   = PREFIX + "WeaponStatus"; //$NON-NLS-1$
     /** The field ID for when the advantage gets Modifiers. */
     public static final String         ID_MODIFIER_STATUS_CHANGED = PREFIX + "Modifier"; //$NON-NLS-1$
+    /** The field ID for user description changes. */
+    public static final String         ID_USER_DESC               = PREFIX + "UserDesc"; //$NON-NLS-1$
     /** The type mask for mental advantages. */
     public static final int            TYPE_MASK_MENTAL           = 1 << 0;
     /** The type mask for physical advantages. */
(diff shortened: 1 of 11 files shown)


crnormand:
Updates:

User Description is now a tab instead of a button/dialog box.
Showing User Description as a tooltip over the Advantage description column is a sheet preference.
DESCRIPTION_USER (and DESCRIPTION_USER_FORMATTED) added to TextOutput.

Unfortunately, I cannot seem to clear the merge conflict.   There is no conflict.   The only difference is the number of spaces before the "=".    Can you still accept the pull request with the conflict?

richardwilkes:
Sorry, just realized this was sitting here waiting on me. I'll try to take a look a bit later today and see what the issue is.

crnormand:
Lol.  No worries.   I’m on the beach in Florida this week 😎



Chris Normand, sent from my iPhone

crnormand:
Oh, I see it... I didn't scroll down far enough.   I just got back, a little more sunburned than expected, so it may take me a day or two to fix this.

crnormand:
Ok, I am at a loss.   The conflict page shows 2 areas of conflict.   One that is just spacing, and the other shows two different instance variables, however, in the PR, the code has BOTH instance variables.   I don't see why this is a conflict.

richardwilkes:
Fixed in a separate PR

Issue #97, opened 2019-01-24
[Feature Request] Alternate Damage Scheme (Thr = Sw-2)


Sorenant:
One often suggested alternative to Know Your Own Strength is to just alter Thrust damage so it's equal to Swing-2, no modification to Basic Lifting and trait costs required. Would it be possible to implement another option just like the KYOS one on Preferences page but without the alteration to BL progression?

PR #124, opened 2019-07-26
Added Thr=SW-2 damage option

crnormand:
I believe this solves https://github.com/richardwilkes/gcs/issues/97.  Since they were requesting "no modification to Basic Lifting and trait costs", it seemed simple enough.

It adds a preference option "Use optional Thrust damage (Thrust = Swing - 2)", and then just redirects the getThrust() method to the getSwing() method, minus 2.

I tested the notifications (although, I must admit, I was blindly copying code), but it seems to work correctly.   If you change the option, you will see the Thr damage change, and you will see all of the weapon damages change (based on thrust).

### DIFF
--- a/src/com/trollworks/gcs/character/CharacterSheet.java
+++ b/src/com/trollworks/gcs/character/CharacterSheet.java
@@ -242,7 +242,7 @@ public class CharacterSheet extends JPanel implements ChangeListener, Scrollable
         if (!GraphicsUtilities.inHeadlessPrintMode()) {
             setDropTarget(new DropTarget(this, this));
         }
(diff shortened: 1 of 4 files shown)


richardwilkes:
In general, this looks reasonable. How does it interact with the other options that modify damage, though? Seems like KYOS and this might be mutually exclusive (although I haven't looked close enough to know for sure). If so, then you shouldn't be able to enable them at the same time.

crnormand:
As I have implemented it, if you turn this feature on, it will redirect the getThrust() query to getSwing() before processing any other thrust damage options.   I have not modified the getSwing() method, so if other strength or swing damage related options are turned on, they will affect the swing damage (and therefore the thrust damage).

It should be completely compatible with any combination of options, assuming the user understands that Thrust is just Swing -2, so there are no separate calculations (or modifiers) for thrust damage.  You could say it is an "override" for Thrust damage.

PR #131, opened 2019-11-15
Note that Go is now required to build

Hubok:
Just noticed this dependency was missing and added some documentation quick.


richardwilkes:
Thanks!

Issue #39, opened 2018-01-15
Support for Ritual Magic (spells as techniques)


richardwilkes:
From [R. William Morris](mailto:noctifer1011@gmail.com):

I use these rules for Clerics in my own GURPS Fantasy game and it'd be nice to be able to create spells that use the Ritual Magic alternative rules (B242 and Thau72).  Currently, I'm 'hacking' by creating custom techniques that are spells, but that leaves me without all the wonderful number-crunching that goes along with the standard spells.

Personally, I'd be fine with just the capability of creating my own spells using Techniques, not necessarily including a copy of each spell as a Technique in the install package.

Maeriden:
I'm interested in this feature as well.
I could try to do it myself, but I dont know how to integrate it or which files to look at. Would you be willing to help me? @richardwilkes

richardwilkes:
This requires code changes that may or may not be trivial -- I haven't looked into it. Unfortunately, I don't have a lot of spare time right now and this is unlikely to be worked on by me any time soon.

Maeriden:
The following is mostly speculation since I havent really looked at much of the code, but maybe it can serve as a starting point.
I figured this can be done in two ways: adding a separate "SpellTechnique" class or modify the existing Spell class.

Adding a separate class would offer more flexibility but probably require more extensive work to be integrated with the rest of the code.
It would also require a new spell library, but I think it can be generated programmatically from the existing ones.

Modifing the existing class would require adding an "isTechnique" and (maybe, see below) a "prerequisitePenalty" fields. They could probably be fitted beside the "Duration" field in the details editor.
If a spell is marked as a technique

- Its "College" field would be the specialization of a "College" skill (e.g. Fireball belongs to the Fire college so it is associated with the skill "College (Fire)").
  - This is not a prerequisite, since Thaumatology 72 sas that ritual mages can cast spells even if they dont have points in the college skill.
  - College skill level should be capped at the "core skill" level.
- Its difficulty would be locked to H, or could be left editable for flexibility.
- Its controlling attribute would become meaningless (maybe make it non-editable and gray it out).
- Its default level would be \<college skill\> + \<prerequisite penalty\>.
  - One problem with this is that the "core skill" for magic, to which college skills default, is not fixed (the manuals suggest "Ritual Magic" or "Thaumatology"), so if a user added a spell but not the college skill (because he has not spent points on it), we cant compute the spell level.
[...]

PR #165, opened 2020-04-01
Support for Ritual Magic (spells as techniques)

Maeriden:
Since my ritual mage got the ability to cast spells last week I decided to tackle https://github.com/richardwilkes/gcs/issues/39.
I don't really like how the code turned out but I wanted to have something that works rather than not have anything at all. Most of it is a mix of copy-pasted code from Spell and Technique. I did a couple of quick tests and it seems to work ok.
Do let me know if there is anything bad.

### DIFF
--- a/src/com/trollworks/gcs/spell/Spell.java
+++ b/src/com/trollworks/gcs/spell/Spell.java
@@ -106,8 +106,8 @@ public class Spell extends ListRow implements HasSourceReference {
     private              String            mMaintenance;
     private              String            mCastingTime;
     private              String            mDuration;
-    private              int               mPoints;
-    private              SkillLevel        mLevel;
+    protected            int               mPoints;
+    protected            SkillLevel        mLevel;
     private              SkillAttribute    mAttribute;
     private              String            mReference;
     private              boolean           mIsVeryHard;
--- a/src/com/trollworks/gcs/spell/Spell.java
+++ b/src/com/trollworks/gcs/spell/Spell.java
@@ -19,6 +19,7 @@ import com.trollworks.gcs.common.ListFile;
 import com.trollworks.gcs.common.LoadState;
 import com.trollworks.gcs.skill.SkillAttribute;
 import com.trollworks.gcs.skill.SkillDefault;
+import com.trollworks.gcs.skill.SkillDifficulty;
 import com.trollworks.gcs.skill.SkillLevel;
 import com.trollworks.gcs.weapon.MeleeWeaponStats;
 import com.trollworks.gcs.weapon.RangedWeaponStats;
@@ -110,7 +111,7 @@ public class Spell extends ListRow implements HasSourceReference {
     protected            SkillLevel        mLevel;
     private              SkillAttribute    mAttribute;
     private              String            mReference;
-    private              boolean           mIsVeryHard;
+    private              SkillDifficulty   mDifficulty;
     private              List<WeaponStats> mWeapons;
 
     /**
@@ -133,7 +134,7 @@ public class Spell extends ListRow implements HasSourceReference {
         mDuration = isContainer ? "" : getDefaultDuration();
         mPoints = 1;
         mReference = "";
-        mIsVeryHard = false;
+        mDifficulty = SkillDifficulty.H;
         mWeapons = new ArrayList<>();
         updateLevel(false);
     }
@@ -160,7 +161,7 @@ public class Spell extends ListRow implements HasSourceReference {
         mDuration = spell.mDuration;
         mPoints = forSheet ? spell.mPoints : 1;
         mReference = spell.mReference;
-        mIsVeryHard = spell.mIsVeryHard;
+        mDifficulty = spell.mDifficulty;
         if (forSheet && dataFile instanceof GURPSCharacter) {
             if (mTechLevel != null) {
                 mTechLevel = ((GURPSCharacter) dataFile).getDescription().getTechLevel();
@@ -207,7 +208,7 @@ public class Spell extends ListRow implements HasSourceReference {
         }
         if (obj instanceof Spell && super.isEquivalentTo(obj)) {
             Spell row = (Spell) obj;
-            if (mIsVeryHard == row.mIsVeryHard && mPoints == row.mPoints && mLevel.isSameLevelAs(row.mLevel) && mAttribute == row.mAttribute) {
+            if (mDifficulty == row.mDifficulty && mPoints == row.mPoints && mLevel.isSameLevelAs(row.mLevel) && mAttribute == row.mAttribute) {
                 if (Objects.equals(mTechLevel, row.mTechLevel)) {
                     if (mName.equals(row.mName) && mCollege.equals(row.mCollege) && mPowerSource.equals(row.mPowerSource) && mSpellClass.equals(row.mSpellClass) && mReference.equals(row.mReference)) {
                         if (mCastingCost.equals(row.mCastingCost) && mMaintenance.equals(row.mMaintenance) && mCastingTime.equals(row.mCastingTime) && mDuration.equals(row.mDuration)) {
@@ -261,14 +262,14 @@ public class Spell extends ListRow implements HasSourceReference {
         mDuration = isContainer ? "" : getDefaultDuration();
         mPoints = 1;
         mReference = "";
-        mIsVeryHard = false;
+        mDifficulty = SkillDifficulty.H;
         mWeapons = new ArrayList<>();
     }
 
     @Override
     protected void loadAttributes(XMLReader reader, LoadState state) {
         super.loadAttributes(reader, state);
-        mIsVeryHard = reader.isAttributeSet(ATTRIBUTE_VERY_HARD);
+        mDifficulty = reader.isAttributeSet(ATTRIBUTE_VERY_HARD) ? SkillDifficulty.VH : SkillDifficulty.H;
     }
 
     @Override
@@ -279,7 +280,7 @@ public class Spell extends ListRow implements HasSourceReference {
             // Fix for legacy format...
             if (mName.toLowerCase().endsWith("(vh)")) {
                 mName = mName.substring(0, mName.length() - 4).trim();
-                mIsVeryHard = true;
+                mDifficulty = SkillDifficulty.VH;
             }
         } else if (TAG_TECH_LEVEL.equals(name)) {
             mTechLevel = reader.readText();
@@ -332,7 +333,7 @@ public class Spell extends ListRow implements HasSourceReference {
 
     @Override
     protected void saveAttributes(XMLWriter out, boolean forUndo) {
-        if (mIsVeryHard) {
+        if (isVeryHard()) {
             out.writeAttribute(ATTRIBUTE_VERY_HARD, true);
         }
     }
@@ -420,7 +421,7 @@ public class Spell extends ListRow implements HasSourceReference {
 
     /** @return The calculated spell skill level. */
     private SkillLevel calculateLevelSelf() {
-        return calculateLevel(getCharacter(), mPoints, mAttribute, mIsVeryHard, mCollege, mPowerSource, mName, getCategories());
+        return calculateLevel(getCharacter(), mPoints, mAttribute, isVeryHard(), mCollege, mPowerSource, mName, getCategories());
     }
 
     /**
@@ -728,7 +729,12 @@ public class Spell extends ListRow implements HasSourceReference {
 
     /** @return Whether this is a "Very Hard" spell or not. */
     public boolean isVeryHard() {
-        return mIsVeryHard;
+        return mDifficulty == SkillDifficulty.VH;
+    }
+
+    /** @return The difficulty. */
+    public SkillDifficulty getDifficulty() {
+        return mDifficulty;
     }
 
     /**
@@ -737,9 +743,9 @@ public class Spell extends ListRow implements HasSourceReference {
      * @return Whether it was modified.
      */
     public boolean setDifficulty(SkillAttribute attribute, boolean veryHard) {
-        if (mAttribute != attribute || mIsVeryHard != veryHard) {
+        if (mAttribute != attribute || isVeryHard() != veryHard) {
             mAttribute = attribute;
-            mIsVeryHard = veryHard;
+            mDifficulty = veryHard ? SkillDifficulty.VH : SkillDifficulty.H;
             startNotify();
             notify(ID_DIFFICULTY, this);
             updateLevel(true);
--- a/src/com/trollworks/gcs/spell/Spell.java
+++ b/src/com/trollworks/gcs/spell/Spell.java
@@ -271,8 +271,9 @@ public class Spell extends ListRow implements HasSourceReference {
     protected void loadAttributes(XMLReader reader, LoadState state) {
         super.loadAttributes(reader, state);
         // Compatibility with version 3 of the Spell format
-        if (reader.isAttributeSet(ATTRIBUTE_VERY_HARD))
+        if (reader.isAttributeSet(ATTRIBUTE_VERY_HARD)) {
             mDifficulty = SkillDifficulty.VH;
+        }
     }
 
     @Override
--- a/src/com/trollworks/gcs/spell/Spell.java
+++ b/src/com/trollworks/gcs/spell/Spell.java
@@ -44,7 +44,7 @@ import java.util.Set;
 
 /** A GURPS Spell. */
 public class Spell extends ListRow implements HasSourceReference {
-    private static final int               CURRENT_VERSION          = 3;
+    private static final int               CURRENT_VERSION          = 4;
     /** The extension for Spell lists. */
     public static final  String            OLD_SPELL_EXTENSION      = "spl";
     /** The XML tag used for items. */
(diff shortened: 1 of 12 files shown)


richardwilkes:
Thanks, @Maeriden -- I'll try to find some time to review this and provide any feedback over the weekend.

src/com/trollworks/gcs/spell/SpellTechnique.java
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    // Copied from Spell class
+    private static int getSpellBonuses(GURPSCharacter character, String id, String qualifier, Set<String> categories, StringBuilder toolTip) {
+        int level = character.getIntegerBonusFor(id, toolTip);
+        level += character.getIntegerBonusFor(id + '/' + qualifier.toLowerCase(), toolTip);
+        level += character.getSpellComparedIntegerBonusFor(id + '*', qualifier, categories, toolTip);
+        return level;
+    }
```
richardwilkes, line 124: Don't copy the code, just make the one in Spell.java accessible and call that.

src/com/trollworks/gcs/spell/SpellTechnique.java
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @param builder The {@link StringBuilder} to append this technique's satisfied/unsatisfied
+     *                description to. May be {@code null}.
+     * @param prefix  The prefix to add to each line appended to the builder.
+     * @return {@code true} if this technique has its default satisfied.
+     */
+    public boolean satisfied(StringBuilder builder, String prefix) {
+        if(getCollege() == null || getCollege().isBlank()) {
```
richardwilkes, line 176: Please move the common call to `getCollege()` out to a variable so that it only gets called once

src/com/trollworks/gcs/spell/SpellTechnique.java
```diff
@@ -0,0 +1,214 @@
...
+        return mSpellPrerequisiteCount;
+    }
+
+    public boolean setSpellPrerequisiteCount(int spellPrerequisiteCount) {
+        if (mSpellPrerequisiteCount != spellPrerequisiteCount) {
+            mSpellPrerequisiteCount = spellPrerequisiteCount;
+            return true;
+        }
+        return false;
+    }
```
richardwilkes, line 213: Seems like this should really be a calculated value, based on actual prereqs.

src/com/trollworks/gcs/spell/SpellTechnique.java
```diff
@@ -0,0 +1,214 @@
...
+import com.trollworks.toolkit.io.xml.XMLReader;
+import com.trollworks.toolkit.utility.I18n;
+
+import java.io.IOException;
+import java.text.MessageFormat;
+import java.util.HashSet;
+import java.util.Set;
+
+/** A GURPS Spell for the Ritual Magic system. */
+public class SpellTechnique extends Spell {
```
richardwilkes, line 35: Since these seem to be named "Ritual Magic Spell" or something similar throughout the interface, the class should and xml tags should probably reflect that as well

src/com/trollworks/gcs/spell/Spell.java
```diff
@@ -261,14 +262,14 @@ protected void prepareForLoad(LoadState state) {
...
-        mIsVeryHard = false;
+        mDifficulty = SkillDifficulty.H;
         mWeapons = new ArrayList<>();
     }
 
     @Override
     protected void loadAttributes(XMLReader reader, LoadState state) {
         super.loadAttributes(reader, state);
-        mIsVeryHard = reader.isAttributeSet(ATTRIBUTE_VERY_HARD);
+        mDifficulty = reader.isAttributeSet(ATTRIBUTE_VERY_HARD) ? SkillDifficulty.VH : SkillDifficulty.H;
```
richardwilkes, line 272: If you're going to switch this to use the SkillDifficulty enum, then it has to support all possible values, not just H/VH. In addition, this should no longer be an attribute and should be stored like it is for skills. When loading old versions, the attribute needs to be looked at, but for the new version, only the new tag should be considered.

src/com/trollworks/gcs/spell/Spell.java
```diff
@@ -332,7 +333,7 @@ protected void finishedLoading(LoadState state) {
 
     @Override
     protected void saveAttributes(XMLWriter out, boolean forUndo) {
-        if (mIsVeryHard) {
+        if (isVeryHard()) {
             out.writeAttribute(ATTRIBUTE_VERY_HARD, true);
         }
     }
```
richardwilkes, line 339: See note about difficulty above -- this should go away in the new code.

src/com/trollworks/gcs/spell/Spell.java
```diff
@@ -728,7 +729,12 @@ public SkillAttribute getAttribute() {
 
     /** @return Whether this is a "Very Hard" spell or not. */
     public boolean isVeryHard() {
-        return mIsVeryHard;
+        return mDifficulty == SkillDifficulty.VH;
+    }
```
richardwilkes, line 733: `isVeryHard()` and all of its uses needs to be removed and replaced with checking the new difficulty value instead

src/com/trollworks/gcs/spell/Spell.java
```diff
@@ -737,9 +743,9 @@ public boolean isVeryHard() {
      * @return Whether it was modified.
      */
     public boolean setDifficulty(SkillAttribute attribute, boolean veryHard) {
-        if (mAttribute != attribute || mIsVeryHard != veryHard) {
+        if (mAttribute != attribute || isVeryHard() != veryHard) {
             mAttribute = attribute;
-            mIsVeryHard = veryHard;
+            mDifficulty = veryHard ? SkillDifficulty.VH : SkillDifficulty.H;
```
richardwilkes, line 748: Shouldn't constrain it to H/VH anymore

src/com/trollworks/gcs/spell/SpellTechniqueEditor.java
```diff
@@ -0,0 +1,415 @@
...
+    private JTextField         mDefaultModifierField;
+    private JTextField         mLevelField;
+    private JTextField         mReferenceField;
+    private JTextField         mNotesField;
+    private JTextField         mCategoriesField;
+    private JTabbedPane        mTabPanel;
+    private PrereqsPanel       mPrereqs;
+    private MeleeWeaponEditor  mMeleeWeapons;
+    private RangedWeaponEditor mRangedWeapons;
+
```
richardwilkes, line 78: Please remove the empty line

src/com/trollworks/gcs/spell/SpellTechniqueEditor.java
```diff
@@ -0,0 +1,415 @@
...
+import javax.swing.JPanel;
+import javax.swing.JScrollPane;
+import javax.swing.JTabbedPane;
+import javax.swing.JTextField;
+import javax.swing.SwingConstants;
+import javax.swing.event.DocumentEvent;
+import javax.swing.event.DocumentListener;
+import javax.swing.text.Document;
+
+public class SpellTechniqueEditor extends RowEditor<SpellTechnique> implements ActionListener, DocumentListener {
```
richardwilkes, line 56: It appears that most of this file is copied from SpellEditor.java. Can a common base class be extracted so we don't have so much duplicated code that will need to be touched up in multiple places in the future?

src/com/trollworks/gcs/spell/SpellTechniqueEditor.java
```diff
@@ -0,0 +1,415 @@
...
+        } else if (doc == mClassField.getDocument()) {
+            LinkedLabel.setErrorMessage(mClassField, mClassField.getText().trim().isEmpty() ? I18n.Text("The class field may not be empty") : null);
+        } else if (doc == mClassField.getDocument()) {
+            LinkedLabel.setErrorMessage(mCastingCostField, mCastingCostField.getText().trim().isEmpty() ? I18n.Text("The casting cost field may not be empty") : null);
+        } else if (doc == mClassField.getDocument()) {
+            LinkedLabel.setErrorMessage(mCastingTimeField, mCastingTimeField.getText().trim().isEmpty() ? I18n.Text("The casting time field may not be empty") : null);
+        } else if (doc == mClassField.getDocument()) {
+            LinkedLabel.setErrorMessage(mDurationField, mDurationField.getText().trim().isEmpty() ? I18n.Text("The duration field may not be empty") : null);
+        }
+
```
richardwilkes, line 235: Please remove blank line

richardwilkes:
Oh, and please make sure you run the formatter across the changed code. Assuming you're using IntelliJ, the formatting options that I want should already be setup in the https://github.com/richardwilkes/gcs_intellij repo

src/com/trollworks/gcs/spell/SpellTechnique.java
Maeriden: Thaumatology 72-73 talks about a few issues relative to determining prerequisite count

- How to handle non-spell prerequisites. Options include satisfying them, ignoring them, or making them add a penalty of 1 per 10 full points of their cost.
- Double-counting of prerequisites. If spell D depends on spells B and C, and both depend on spell A (diamond pattern), is the prereq count 3 or 4? (luckily the book suggests 3 but does not force it)
- Some spells like Enchant have crazy prereqs (10 other spells) so the manual says the GM might apply a custom penalty to avoid discouraging players from using it.

So I thought it was best to just store the penalty and let the user modify it as he likes.
If this PR is accepted I was thinking of making a new spell library (or add to the existing one) with the RM variant of all the spells in Magic, with a default prereq count based on the table in Thaumatology 261-267.

What do you think?

src/com/trollworks/gcs/spell/SpellTechnique.java
Maeriden: Would SpellRitualMagic, SpellRitualMagicEditor be ok?

src/com/trollworks/gcs/spell/Spell.java
Maeriden: Alright, I wasn't sure how to handle backwards compatibility. I assume this would increase the version of the Spell class, yes?

src/com/trollworks/gcs/spell/SpellTechniqueEditor.java
Maeriden: I guess it can. My approach was to touch as little external code as possible to avoid fucking something up.
I'll see what I can come up with.

src/com/trollworks/gcs/spell/SpellTechnique.java
richardwilkes: OK, sounds reasonable to have it just be a input field, then.

As for a new spell list, yes, I think that's best. Probably best to create a new libraries with a " (Ritual Magic)" suffix, i.e. "Magic (Ritual Magic)", etc.

src/com/trollworks/gcs/spell/SpellTechnique.java
richardwilkes: I'd prefer `RitualMagicSpell`, `RitualMagicSpellEditor`, etc. so that it reads naturally.

src/com/trollworks/gcs/spell/Spell.java
richardwilkes: Yes

richardwilkes: You only need to be able to read old versions. The code should only ever write the latest version.

src/com/trollworks/gcs/spell/SpellTechniqueEditor.java
richardwilkes: Understandable... but long-term maintenance is important.

richardwilkes:
By the way, you're welcome to join the [GURPS Discord](https://discord.gg/Sv4Svce) and initiate a chat with me (Rich#7409) if you have questions and want to get them answered more interactively. Our hours may not align (I'm in the Pacific Time Zone -- GMT-8, I think), but I sometimes up at odd hours like tonight.

Maeriden:
Sounds good. I'm GMT+1, but having another communication channel can be useful.

About the formatter, I didn't know it was there or I would have definetly used it. I'll make a dedicated commit once all the issues are sorted out.

Maeriden:
First of all, I'm sorry if the commits are a mess, I get sidetracked a lot when doing stuff and when I notice it it's already too late.
Regarding the formatter, I ran it with the "optimize imports" and "cleanup code" options. Rearranging members seemed to modify a lot of files I didn't touch so I assumed you don't use that option.
There are a couple of TODO and FIXME in RitualMagicSpell:

1. I use `Technique.calculateTechniqueLevel()` to get an initial value for the spell level. It has a `specialization` parameter that I'm not sure how to use. `Technique.calculateLevelSelf()` passes in the full name of the default skill, but then I don't get the point of the `def` parameter. If it is indeed intended to be the default skill, I'd say I find the parameter name misleading.
2. Technique.calculateTechniqueLevel()` adds the default skill modifier to the returned level, but not to the relative level. The modifier is instead applied inside `SkillColumn.RELATIVE_LEVEL.getRelativeLevel()` but I don't understand why. Isn't the modifier part of the relative level?
3. I'm unsure how to handle bonuses to spell colleges inside `RitualMagicSpell.calculateLevel()`. Since Ritual Magic uses college skills, a bonus to a college should probably be applied to the College skills, which would then propagate to all the spells that default to that college. The problem with this is the need to convert everything that gives bonuses to spells-of-college (e.g. Magery) to instead give bonuses to college skills. Or maybe I'm completely wrong and this type of bonus does not apply to college skills but directly to spells. I'm not even sure it matters since I don't know if you're ever required to roll a pure college skill. Anyway, a possible solution would be to make a `RitualMagicCollege` class that handles this special case, but I'm not even sure it's possible.
4. I'm not sure how to trigger the update of the spell level when the prerequisite count is changed. What I do now is just to call `updateLevel()` and it seems to work. Since setting the prerequisite count means you are changing the modifier of the default skill, the equivalent operation in the `Technique` class is `setDefault()`, but that method doesn't seem to do anything special to trigger updates.
5. There was something else too but I'm super tired and forgot what it was while writing. Too bad.

On a separate note, I'd like to suggest adding an `isSatisfied()` method to the `Row` class and call that in `PrerequisitesThread.processRows()`. In my opinion it would be cleaner than adding a chain of `if` for classes with special satisfaction requirements.

richardwilkes:
> I use Technique.calculateTechniqueLevel() to get an initial value for the spell level. It has a specialization parameter that I'm not sure how to use. Technique.calculateLevelSelf() passes in the full name of the default skill, but then I don't get the point of the def parameter.
If it is indeed intended to be the default skill, I'd say I find the parameter name misleading.

The issue here is that Techniques (and now Ritual Magic) have specializations that refer to a full skill name + specialization. For a normal skill, it would just be the specialization. Since its a sub-class, the naming is not quite suitable, but that's the way it is...

richardwilkes:
I noticed that you don't actually load saved ritual magic data.

richardwilkes:
> Technique.calculateTechniqueLevel() adds the default skill modifier to the returned level, but not to the relative level. The modifier is instead applied inside SkillColumn.RELATIVE_LEVEL.getRelativeLevel() but I don't understand why. Isn't the modifier part of the relative level?

Good observation. I'm not sure why that was done that way. Probably something to do with some difference in the way skills and techniques are handled in the rules, or possibly just organic growth of the code base. At any rate, it is working as-is and I'm reluctant to go in and change it right now.

richardwilkes:
> I'm unsure how to handle bonuses to spell colleges inside RitualMagicSpell.calculateLevel(). Since Ritual Magic uses college skills, a bonus to a college should probably be applied to the College skills, which would then propagate to all the spells that default to that college.
The problem with this is the need to convert everything that gives bonuses to spells-of-college (e.g. Magery) to instead give bonuses to college skills.
Or maybe I'm completely wrong and this type of bonus does not apply to college skills but directly to spells.
I'm not even sure it matters since I don't know if you're ever required to roll a pure college skill.
Anyway, a possible solution would be to make a RitualMagicCollege class that handles this special case, but I'm not even sure it's possible.

I've never used the Ritual Magic stuff, so couldn't say. I'm unwilling to have a spell modifier also modify a skill, though. If Ritual Magic needs that, the appropriate things giving those modifiers need to have additional features added to explicitly give those bonuses to skills as well.

richardwilkes:
> I'm not sure how to trigger the update of the spell level when the prerequisite count is changed. What I do now is just to call updateLevel() and it seems to work.
Since setting the prerequisite count means you are changing the modifier of the default skill, the equivalent operation in the Technique class is setDefault(), but that method doesn't seem to do anything special to trigger updates.

Testing this in the editor seems to fail... but it also fails on the unmodified code base for skills, apparently. I'll look into fixing that separately, as it used to work.

richardwilkes:
> On a separate note, I'd like to suggest adding an isSatisfied() method to the Row class and call that in PrerequisitesThread.processRows(). In my opinion it would be cleaner than adding a chain of if for classes with special satisfaction requirements.

Yeah, that might be warranted at this point. I don't think it originally applied to enough things to make that call, but nowadays, most everything allows prereqs, so that would likely be better.

richardwilkes:
So... Overall, the changes seem to work well. There are a few rough spots (loading saved data, updating the display in the editors), but I'll look at fixing those up myself.

Related to that, I'm also going to change the name of the skill it looks for from "College" to "Ritual Magic College" to be explicit.

Thanks for taking the time to add this support -- I know a lot of folks have been waiting a long time for me to do something about it, so they will be quite happy when the next release comes out and your changes are made available to them.

PR #205, opened 2020-06-01
Added automatic Windows logging support

Hubok:
Errors on Windows are saved to %localappdata%\GCS\error.log if GCS_LOG env variable is not defined. Includes some code review from Rich in commits https://github.com/richardwilkes/gcs/commit/2899b094bc215bc32692c58158ee6a8a17088bba (comments) and https://github.com/richardwilkes/gcs/commit/e093e124b6b6b5b3be6cd9f9be401c69b8c85f17 (resolutions).

### DIFF
--- a/com.trollworks.gcs/src/com/trollworks/gcs/utility/Log.java
+++ b/com.trollworks.gcs/src/com/trollworks/gcs/utility/Log.java
@@ -31,13 +31,16 @@ public class Log {
             } catch (Throwable throwable) {
                 error("Unable to redirect log to " + property, throwable);
             }
-        } else if (Platform.isWindows() && System.getenv() != null) {
-            File logDirectory = new File(System.getenv("LOCALAPPDATA") + "\\GCS");
-            try {
-                logDirectory.mkdirs();
-                OUT = new PrintStream(logDirectory+"\\error.log");
-            } catch (Throwable throwable) {
-                error("Unable to redirect log to " + logDirectory+"\\error.log", throwable);
+        } else if (Platform.isWindows()) {
+            String localAppData = System.getenv("LOCALAPPDATA");
+            if (localAppData != null) {
+                File logDirectory = new File(System.getenv("LOCALAPPDATA") + "\\GCS");
+                try {
+                    logDirectory.mkdirs();
+                    OUT = new PrintStream(logDirectory + "\\error.log");
+                } catch (Throwable throwable) {
+                    error("Unable to redirect log to " + logDirectory + "\\error.log", throwable);
+                }
             }
         }
     }
--- a/com.trollworks.gcs/src/com/trollworks/gcs/utility/Log.java
+++ b/com.trollworks.gcs/src/com/trollworks/gcs/utility/Log.java
@@ -31,17 +31,13 @@ public class Log {
             } catch (Throwable throwable) {
                 error("Unable to redirect log to " + property, throwable);
             }
-        }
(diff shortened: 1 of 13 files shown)


com.trollworks.gcs/src/com/trollworks/gcs/utility/Log.java
```diff
@@ -30,6 +31,14 @@
             } catch (Throwable throwable) {
                 error("Unable to redirect log to " + property, throwable);
             }
+        } else if (Platform.isWindows() && System.getenv() != null) {
```
richardwilkes, line 34: I think I wasn’t clear before... something like this instead:
```
} else if (Platform.isWindows()) {
    String localAppData = System.getenv(“LOCALAPPDATA”);
    if (localAppData != null) {
        File logDirectory = new File(localAppData + “\\GCS”;
```

Hubok: Fair point, I realize now after I wrote it that I was calling an empty System.getenv() and would not be properly checking for null. Your change also removes an excess call, and that's reflected in my new commit below.

com.trollworks.gcs/src/com/trollworks/gcs/utility/Log.java
```diff
@@ -31,13 +31,16 @@
...
-            File logDirectory = new File(System.getenv("LOCALAPPDATA") + "\\GCS");
-            try {
-                logDirectory.mkdirs();
-                OUT = new PrintStream(logDirectory+"\\error.log");
-            } catch (Throwable throwable) {
-                error("Unable to redirect log to " + logDirectory+"\\error.log", throwable);
+        } else if (Platform.isWindows()) {
+            String localAppData = System.getenv("LOCALAPPDATA");
+            if (localAppData != null) {
+                File logDirectory = new File(System.getenv("LOCALAPPDATA") + "\\GCS");
```
richardwilkes, line 37: Please use `localAppData` instead of calling `System.getenv()` a second time.

Hubok: That's embarrassing, I can't make an excuse for that one. See my amended commit.

richardwilkes:
Hmm... I see there was a new commit, but it doesn't show as having any changes.

Hubok:
There we go, that fixed it. Made a disgusting amount of weird commits but it's better than hard resets. I'll be sure to look into why that happened.

com.trollworks.gcs/src/com/trollworks/gcs/utility/Log.java
```diff
@@ -30,6 +31,17 @@
             } catch (Throwable throwable) {
                 error("Unable to redirect log to " + property, throwable);
             }
+        } else if (Platform.isWindows()) {
+            String localAppData = System.getenv("LOCALAPPDATA");
+            if (localAppData != null) {
+                File logDirectory = new File(localAppData);
```
richardwilkes, line 37: You dropped the `+ "\\GCS"` part, which should either still be there, or the `+ "\\error.log"` cases below should be revised to something like `+ "\\gcs_error.log"`.

Hubok: You have the patience of a saint, thank you. I applied an theme to IntelliJ and it tries to mark what code does inline, I'm gonna go turn that off so I don't make that mistake again.

PR #231, opened 2020-07-09
Mark Will in attribute listing with special value which will be replaced

JochenLinnemann:
I know this is rather "dirty"...

I am working on a German translation and "Will" translates to "Willenskraft". It is used in the attribute listing and for RSL. By marking one with another string I can have the long word in the attribute listing and e.g. WI in the relative skill level.

IMO, it would be best to mark calls to **I18n.Text("Will")** with some kind of (optional) context, e.g. **I18n.Text("Will", "attribute")**. This will require adjusting the Bundler. I could do that if you prefer it.

### DIFF
--- a/com.trollworks.gcs/src/com/trollworks/gcs/character/AttributesPanel.java
+++ b/com.trollworks.gcs/src/com/trollworks/gcs/character/AttributesPanel.java
@@ -38,7 +38,7 @@ public class AttributesPanel extends DropPanel {
         addLabelAndField(sheet, GURPSCharacter.ID_INTELLIGENCE, I18n.Text("Intelligence (IQ)"), true);
         addLabelAndField(sheet, GURPSCharacter.ID_HEALTH, I18n.Text("Health (HT)"), true);
         addDivider();
-        addLabelAndField(sheet, GURPSCharacter.ID_WILL, I18n.Text("Will"), true);
+        addLabelAndField(sheet, GURPSCharacter.ID_WILL, I18n.Text("Will (Will)").replace(" (Will)", ""), true);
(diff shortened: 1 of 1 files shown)


richardwilkes:
Not a fan of this. I take it just using the shortened "Will" for "Willenskraft" everywhere isn't acceptable? If not, I'll have to give some thought to how I would solve this -- your suggestion to add another parameter is likely the route I'd take, but it will need some thought for how the file is read & written to make sure it is understandable to others and doesn't introduce redundancy.

JochenLinnemann:
I could go with "Wille" instead of "Willenskratf".

There is another issue, though: "Portrait". In the context of a character portrait, this is just "Portrait". In the context of printing, this is "Hochformat".

If I were to implement this, I would annotate every location, keep the key just simple in the file, and add the annotation to the value, together with a simple value as a fallback. Like this:

```
k:"Portrait"
v:"Portrait"
v.character:"Portrait"
v.printing:"Hochformat"

k:"Will"
v:"Wille"
v.attribute_listing:"Willenskraft"
v.relative_skill_level:"WI"
```

richardwilkes:
Why is the word different for printing? Note that there is no such distinction in the code -- the widget isn't even aware it is being printed.

richardwilkes:
I just realized you may have been referring to the "Portrait" and "Landscape" text, rather than the portrait widget for the printing case. (sigh)

JochenLinnemann:
Yes, in German there is literal translation:
Portrait = Portrait
Landscape = Landschaft

and contextual translation (here page orientation):
Portrait = Hochformat (literally high format aka vertical format)
Landscape = Querformat (literally cross-format aka horizontal format)

richardwilkes:
Please take a look at PR #233 -- this should provide what you're requesting, albeit not exactly how you were suggesting.

PR #233, opened 2020-07-11
Allow for alternative contexts when loading i18n strings

richardwilkes:
- Use a context with "Portrait" from page setup
- Use a context with "Will" from attributes panel


richardwilkes:
@JochenLinnemann please review this and let me know if you think it will work for your particular use case.

richardwilkes:
A bug was discovered today that I don't want to leave unfixed for long, so I'm going to go ahead and pull this in along with the other changes I've been making. Feel free to comment on it, nonetheless.

JochenLinnemann:
I like that! It is easy enough and straight forward. I will check it with my current dev branch and report any problems.

JochenLinnemann:
In the long run: Is there a way to bundle translation files with the distributable MSI? I could not find anything. Do you have a preferred way for doing this or can I just suggest a solution and provide a PR?

richardwilkes:
I’m a bit reluctant to return to bundling the translations with GCS again. Without an active maintainer, they tend to slip out of sync *very* quickly. Oddly enough, when I used to have German, Russian & Spanish translations bundled, I got a significant number of requests from people to disable them, as they wanted to use English for everything GURPS-related.

So... I’d prefer to keep any available translations as a separately installable thing. That was the intent behind externalizing them a while back and in v4.21 I also added the ability to set an environment variable to control where to look for them, as at least on macOS, modifying the app bundle after deployment can trigger the security on the platform.

richardwilkes:
Note: I would be more than happy to add links to any translations to the GCS web site as well as potentially add some mechanism in GCS itself for looking up and installing external translations somehow.

Issue #28, opened 2018-01-15
Better ordering of modifiers


richardwilkes:
Modifiers are sorted alphabetically when an advantage is copied to the character sheet. This frequently leads to poor ordering.

There also needs to be a way to group modifiers as well as mark groups that can only have the following combinations of items chosen from them:

- No more than one
- Must select one
- Zero or more
- All of this can be done, but it will take a fair amount of effort, reworking both the UI as well as the data model.

richardwilkes:
Modifiers in GCS v5 can be arranged as desired as well put into containers for grouping.

PR #248, opened 2020-08-16
Do not sort advantage and equipment modifiers.

angelforest:
There doesn't appear to be a need to do the sorting as the order is always preserved during saving/loading.
Related to #28


angelforest:
>as the file could have been hand-edited
Which is desirable as certain modifiers are related to each other and thus should be grouped together or ordered in some specific way regardless of their alphabetical order. The most notable offender here is the Enemy disadvantage where power, intent and special modifiers end up being mixed in (seemingly) random order.

richardwilkes:
Unfortunately, there is no way in the UI to reorder modifiers. Until that changes, I'd rather have them sorted automatically.

PR #279, opened 2020-11-11
Add ctrl+tab shortcut to switch current tab of DockContainer

Maeriden:
Adds `ctrl+tab` and `ctrl+shift+tab` as shortcuts to cycle through the tabs of `DockContainer` when it, or any descendent, has focus.

The change motivated by the fact that a lot* of software uses those shortcuts to switch tabs. Among those I have open at the moment, my file explorer, firefox, vscode and intellij have this behavior.

The ctrl modifier is actually chosen using the same code used by menu commands. I hope this makes it cross-platform (i.e. uses ctrl on win and linux, cmd on Mac) but I cannot test it.

As far as I could tell the shortcuts do not interfere with normal focus traversal, except for the loss of ability to change focus even with ctrl held down (but who does that anyway?).

\*A subjective lot

### DIFF
--- a/com.trollworks.gcs/src/com/trollworks/gcs/ui/widget/dock/DockContainer.java
+++ b/com.trollworks.gcs/src/com/trollworks/gcs/ui/widget/dock/DockContainer.java
@@ -14,6 +14,7 @@ package com.trollworks.gcs.ui.widget.dock;
 import com.trollworks.gcs.menu.file.CloseHandler;
 import com.trollworks.gcs.ui.UIUtilities;
 
+import java.awt.AWTKeyStroke;
 import java.awt.Color;
 import java.awt.Component;
 import java.awt.Container;
@@ -22,9 +23,19 @@ import java.awt.EventQueue;
 import java.awt.Insets;
 import java.awt.KeyboardFocusManager;
 import java.awt.LayoutManager;
+import java.awt.Toolkit;
+import java.awt.event.ActionEvent;
+import java.awt.event.InputEvent;
+import java.awt.event.KeyEvent;
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import javax.swing.AbstractAction;
+import javax.swing.ActionMap;
+import javax.swing.InputMap;
 import javax.swing.JPanel;
+import javax.swing.KeyStroke;
 
 /**
  * All {@link Dockable}s are wrapped in a {@link DockContainer} when placed within a {@link Dock}.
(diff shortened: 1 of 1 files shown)


com.trollworks.gcs/src/com/trollworks/gcs/ui/widget/dock/DockContainer.java
```diff
@@ -332,4 +344,72 @@ public void layoutContainer(Container parent) {
             current.setBounds(insets.left, insets.top + height, width, remaining);
         }
     }
+
+    private void setupDockablesSwitchKeys() {
+        // Standard modifier for shortcuts _should_ be CTRL on win and linux and CMD on mac
+        int       standardModifier = Toolkit.getDefaultToolkit().getMenuShortcutKeyMaskEx();
```
richardwilkes, line 350: Don't want to use this on Mac, since Cmd-Tab and Shift-Cmd-Tab is already a system-wide key binding (it is for switching apps via the keyboard). Use of Ctrl on the Mac is sufficient.

com.trollworks.gcs/src/com/trollworks/gcs/ui/widget/dock/DockContainer.java
```diff
@@ -332,4 +344,72 @@ public void layoutContainer(Container parent) {
...
+
+        // Map action names to actions
+        ActionMap actionMap = getActionMap();
+        actionMap.put("ShowNextDockable", new ShowNextDockableAction());
+        actionMap.put("ShowPrevDockable", new ShowPrevDockableAction());
+    }
+
+
+    static class ShowNextDockableAction extends AbstractAction {
+        private static final long serialVersionUID = 1L;
```
richardwilkes, line 376: These serialVersionUID constants aren't needed. Please remove them.

richardwilkes (commented):
Generally looks good. I still need to find a spare moment to grab your branch and try it out locally to verify, but I don't see any issues other than the command key one.

richardwilkes:
OK, gave it a whirl and after changing the code to always use the CTRL key as the standard modifier, it works well.

Maeriden:
That's great.

I forgot to mention, the original plan was to make the shortcuts configurable, but I didn't know how to retrieve keystrokes defined in the preferences from the DockContainers, and how to handle configuration changes, so I opted to make them constant.

richardwilkes:
Given what this does and how it maps to standard usage across platforms, I think its fine not making them configurable. That can always be revisited if a real need comes up.

Thanks for the PR.

richardwilkes (approved):

PR #286, opened 2020-12-23
Ritual Magic Spell Object Ignores Difficulty Change-patch

PlasticPhoenix:
This attempts to fix the Ritual Magic Spell objects ignoring changes to their difficulty inside the Ritual Magic Spell editor.


PlasticPhoenix:
Heh, not that easy, I see.

PR #292, opened 2021-01-30
Stretch encumbrance/lift panel instead of FP/HP panel.

angelforest:
Currently, when trying to fill the entire width of the page, the FP/HP panel is stretched, which results in very wide FP/HP fields. This change makes it so the encumbrance panel fills the empty space instead, which is less noticeable and that's how it was done in the old layout.

### DIFF
--- a/com.trollworks.gcs/src/com/trollworks/gcs/character/SettingsEditor.java
+++ b/com.trollworks.gcs/src/com/trollworks/gcs/character/SettingsEditor.java
@@ -59,7 +59,7 @@ public class SettingsEditor extends BaseWindow implements ActionListener, Docume
     private JCheckBox                mShowCollegeInSpells;
     private JCheckBox                mShowDifficulty;
     private JCheckBox                mShowTitleInsteadOfNameInPageFooter;
-    private JCheckBox                mExtraSpaceAroundFP;
+    private JCheckBox                mExtraSpaceAroundEncumbrance;
     private JComboBox<LengthUnits>   mLengthUnitsCombo;
     private JComboBox<WeightUnits>   mWeightUnitsCombo;
     private JComboBox<DisplayOption> mUserDescriptionDisplayCombo;
@@ -111,7 +111,7 @@ public class SettingsEditor extends BaseWindow implements ActionListener, Docume
         mShowCollegeInSpells = addCheckBox(panel, I18n.Text("Show the College column in the spells list"), null, mSettings.showCollegeInSpells());
         mShowDifficulty = addCheckBox(panel, I18n.Text("Show the Difficulty column in the skills and spells lists"), null, mSettings.showDifficulty());
         mShowTitleInsteadOfNameInPageFooter = addCheckBox(panel, I18n.Text("Show the title rather than the name in the page footer"), null, mSettings.useTitleInFooter());
-        mExtraSpaceAroundFP = addCheckBox(panel, I18n.Text("Add extra space around FP/HP table rather than around Encumbrance table"), null, mSettings.extraSpaceAroundFP());
+        mExtraSpaceAroundEncumbrance = addCheckBox(panel, I18n.Text("Add extra space around Encumbrance table rather than around FP/HP table"), null, mSettings.extraSpaceAroundEncumbrance());
         mBaseWillOn10 = addCheckBox(panel, I18n.Text("Base Will on 10 and not IQ"), null, mSettings.baseWillOn10());
         mBasePerOn10 = addCheckBox(panel, I18n.Text("Base Perception on 10 and not IQ"), null, mSettings.basePerOn10());
         mUseMultiplicativeModifiers = addCheckBox(panel, I18n.Text("Use Multiplicative Modifiers from PW102 (note: changes point value)"), null, mSettings.useMultiplicativeModifiers());
@@ -194,8 +194,8 @@ public class SettingsEditor extends BaseWindow implements ActionListener, Docume
             mSettings.setShowDifficulty(mShowDifficulty.isSelected());
         } else if (source == mShowTitleInsteadOfNameInPageFooter) {
             mSettings.setUseTitleInFooter(mShowTitleInsteadOfNameInPageFooter.isSelected());
-        } else if (source == mExtraSpaceAroundFP) {
-            mSettings.setExtraSpaceAroundFP(mExtraSpaceAroundFP.isSelected());
+        } else if (source == mExtraSpaceAroundEncumbrance) {
+            mSettings.setExtraSpaceAroundEncumbrance(mExtraSpaceAroundEncumbrance.isSelected());
         } else if (source == mBaseWillOn10) {
             mSettings.setBaseWillOn10(mBaseWillOn10.isSelected());
         } else if (source == mBasePerOn10) {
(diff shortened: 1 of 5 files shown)


richardwilkes:
This was changed to put the excess space where it is so that printed sheets got the most space in an area where the player is changing the values on a regular basis. While I agree it doesn't look great when there is a lot of space free, I'm unwilling to make the paper sheet less useable.

angelforest:
I see, that makes sense for printed sheets. Ah, the tribulations of WYSIWYG. Would it be okay to add an option in settings to switch between the two variants?

richardwilkes:
Yes, I'd be fine with that.

angelforest:
All done, I believe.

richardwilkes:
Mostly looks good. However, I'd like the default to be what it currently is, so that all existing sheets stay as-is without explicitly asking for the change. i.e. make the boolean reflect wanting the extra space in the encumbrance table, rather than in the hp/fp table, and default it to false.

A third option that might be appealing as well would be distributing the excess space to both. Not sure if that would be worth the effort.

richardwilkes:
Another consideration... the output templates should probably be adjusted so that they can do the right thing depending on which option was chosen, so that the HTML sheet can match if that's what someone really wants.

angelforest:
>A third option that might be appealing as well would be distributing the excess space to both. Not sure if that would be worth the effort.

It would look about as ugly as the default variant. Doesn't seem to be worth it.

>output templates

It probably should, but currently there is no if-else operator, so there is no way to act on the flag. The only alternative is to output correct css parameter so template can use it directly. Seems hacky to me, but it'll work. Incidentally, templates don't respect "show skill difficulty" setting either, probably for the same reason.

richardwilkes:
> > output templates
> 
> It probably should, but currently there is no if-else operator, so there is no way to act on the flag. The only alternative is to output correct css parameter so template can use it directly. Seems hacky to me, but it'll work. Incidentally, templates don't respect "show skill difficulty" setting either, probably for the same reason.

Good point about the skill difficulty (and probably the spell college column, too). I can’t remember whether I added keys for them or not, but even if I did, there would need to be a way to notice the setting and then add some sort of if/else construct, or in the case of html output, use it to generate some js to do the work.

It is reasonable to bail on the output template support for now.

I’m not near my computer at the moment, but I’ll try to find a moment tomorrow to review the most recent changes and assuming nothing else comes up, merge them in. Thanks!

com.trollworks.gcs/src/com/trollworks/gcs/character/SettingsEditor.java
```diff
@@ -110,6 +111,7 @@ private void addTopPanel() {
         mShowCollegeInSpells = addCheckBox(panel, I18n.Text("Show the College column in the spells list"), null, mSettings.showCollegeInSpells());
```
aimiebell, line 111: Our static analysis engine detects that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

com.trollworks.gcs/src/com/trollworks/gcs/preferences/DisplayPreferences.java
```diff
@@ -70,6 +71,7 @@ public DisplayPreferences(PreferencesWindow owner) {
         mShowCollegeInSheetSpells = addCheckBox(I18n.Text("Show the College column in character sheet spells list *"), prefs.showCollegeInSheetSpells());
```
aimiebell, line 71: Our static analysis engine detects that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

com.trollworks.gcs/src/com/trollworks/gcs/character/SettingsEditor.java
richardwilkes: I am not interested in such changes. Please do not continue to post them.

com.trollworks.gcs/src/com/trollworks/gcs/preferences/DisplayPreferences.java
```diff
@@ -70,6 +71,7 @@ public DisplayPreferences(PreferencesWindow owner) {
...
         mShowDifficulty = addCheckBox(I18n.Text("Show the Difficulty column in character sheet skills and spells lists *"), prefs.showDifficulty());
         mShowTitleInsteadOfNameInPageFooter = addCheckBox(I18n.Text("Show the title rather than the name in the page footer on character sheets *"), prefs.useTitleInFooter());
+        mExtraSpaceAroundEncumbrance = addCheckBox(I18n.Text("Add extra space around Encumrance table rather than around FP/HP table *"), prefs.extraSpaceAroundEncumbrance());
```
richardwilkes, line 74: Spelling: "Encumrance" -> "Encumbrance"

richardwilkes (approved):

com.trollworks.gcs/src/com/trollworks/gcs/character/SettingsEditor.java
aimiebell, line 111: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

com.trollworks.gcs/src/com/trollworks/gcs/preferences/DisplayPreferences.java
aimiebell, line 71: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

PR #309, opened 2021-03-20
Natural Attacks data localization

tiagoaquinofl:
In order to get natural attacks equally translated like others attacks we are able to translate through libraries.

### DIFF
--- a/com.trollworks.gcs/src/com/trollworks/gcs/menu/item/AddNaturalAttacksAdvantageCommand.java
+++ b/com.trollworks.gcs/src/com/trollworks/gcs/menu/item/AddNaturalAttacksAdvantageCommand.java
@@ -104,8 +104,8 @@ public class AddNaturalAttacksAdvantageCommand extends Command {
         bite.setDamage(damage);
         bite.setUsage(I18n.Text("Bite"));
         bite.setReach("C");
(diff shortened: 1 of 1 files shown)


richardwilkes:
With the possible exception of the "usage" field, I'm unwilling to accept localizations for those fields, as they may be used in calculations elsewhere. In general, translations should not be done for the data, as internal calculations sometimes use the english values. Likewise, if a translation is done in a data file, you need to make sure that *all* other data files are also translated, since there are many references between things based on the names used.

Essentially, I don't view translation of the data files a something that can be reasonably done, at least not with GCS's current setup. Language translation must be done at display time only, never in the actual data, in order to be workable.

tiagoaquinofl:
Of course I'm leaving this up to you. As far as we've been translatating the libraries, the "usage", "parry" and "block" (as "No") are used only to display in the "attack" block. I'm not goin any further than this natural attack. But I undestand if you feel that's not safe.

richardwilkes:
I've added the localization support to the "usage" field that is created in this commit: 88de57ce23d30b8aa8e28fceea7d8880d879ebd8

I'm going to go ahead and close this PR since I currently view changing the other fields to have localized data as harmful. Hope the one change is enough.

PR #329, opened 2021-04-23
Add new setting - Phoenixflame's Rescaled Swing damage

Legendsmith:
A middle ground between "Reduced swing for Dungeon fantasy" where swing is just thrust +2, and the RAW Basic Set swing rules where swing can be double thrust, PhoenixFlame's Rescaled Swing damage bases swing on thrust with a per-die bonus and a +1 at even levels of damage, creating a situation where thrust does not gain massive per die bonuses from weapon master and similar traits too early.

I'm pretty sure this should just be able to slot in nicely. It works, doesn't have any conflicts or anything. There's no bugs as far as I know either.


Legendsmith:
Hold off on merging this, I have to check if a particular breakpoint is calculating in the right way with PhoenixFlame.

PR #393, opened 2021-11-03
Add New Damage Progression: Phoenixflame d3 Damage

Legendsmith:
- Adds PhoenixFlame d3 damage progression, which uses d3s.
- Adds options to automatically convert the output (not the data) of d6s to d3s *only* when Phoenix D3 is the chosen damage progression
- Adds a percentage modifier to base sw/thr damage that only appears when Phoenix D3 is active, this is used for the alternate swing damage rules that Phoenix D3 uses.
- Set per die bonuses for d3s (only d3s) when Phoenix D3 is active to only apply every 2 die, these are the rules to prevent damage inflation.  So per die becomes per-every-second-die because 2d3 is roughly equivalent to 1d6. (In the future I might look at how to make +2 per d6 become +1 per d3)

# Goals
1. Decrease wide variability of the d6 for damage
2. Remove the flawed breakpoint of xd+2 to yd-1
3. Every level of Strength is valuable
4. Change the way swing works to provide more viable use of two handed, crushing, and unbalanced weapons

This writeup is 90% Phoenixflame's own words but I've slightly added a few things for clarity.

## Quick Notes
- d6 good for skill.
- d6 bad for damage.
- d3 good for damage.
- swing is now based on thrust but with a percentage modifier
- In the weapon editor, set the percentage bonus of SWING weapons only based on this:
	- Swing Damage reach 1+: 25%
	- Two Handed+25%
	- Crushing +25%
	- Unbalanced +25%
Thus a two handed crushing unbalanced weapon will have a percentage strength bonus of +100%

## Decrease wide variability of the d6 for damage
10 (HP) isn't a lot of resolution, and in a game with damage doing multiplers in increments of D6, it's very much not enough resolution at all.  So we'll need to change some things.  Either HP needs to come up, damage variance needs to be tightened, damage needs to go down, or some combination of these.  Let's take a look.  Things I'm less sure on will be prefixed with a question mark.
First off, damage.  D6 is too much variance, and most normal humans do 1d+X damage, anywhere from -3 to +3.  That *entire range* is the size of the d6 being rolled, and all results are equally likely.  GURPS is a game that doesn't like all results being equally likely, and even DnD has less variance in damage (1d8+16...) even if its results rolls are completely random.

Swing's gone. Instead, swing is a modifier on thrust damage based on weapon used. 
+25% reach 1+ swing
+25% for each of two-handed, crushing or unbalanced.

These numbers are applied to the *basic* strength based damage; the weapon's own modifiers are then applied after. (The implementation reflects this)

So we replace everything with D3s.  D3s have an interesting benefit of being whole number averages (2, rather than 3.5), which allows us to much, much more easily futz with damage modifiers to weapons as well as scaling of strength to damage values.  It also lets us run that ST10 1d6-3 punch damage into something more like 1d3-1 punch damage, which does damage 2/3rds of the time instead of 1/2. (Crushing probably should have a damage floor anyway...)

## Remove the flawed breakpoint of xd+2 to yd-1
In the basic GURPS damage progression, xd+2 to yd-1 is a 0.5 increase where all other increases are a whole number. This is undesirable. 
And, in the general strength progression I find such breakpoints (and in damage calculation to dice in general) undesirable
As 1d+4 = 2d while 1d+7 = 3d
1d+7 does indeed = 3d, as it's 10.5 in both cases
but 1d+4 is 7.5 and 2d is 7. Now one might quibble that a half a point of damage on average isn't a big deal. And, if they're playing high tech, they're probably correct. But when a character by default has 10hp and most injury codes multiply injury, a half point of damage variation being "lost" in a rounding calculation is unacceptable

This is achieved rather easily, basic damage is now a whole die. Damage starts at 1d3 at 10 ST and continues up from there with each level of ST giving +1. This means that every level of strength is now valuable for thrust damage.

## Every level of strength is valuable
(Optionally, but encouraged) Striking strength now costs 10pts/level and gives +1/level to thrust damage with the caveat one should have about half as much SS as lifting strength. Since all muscle damage gets +1 per level now thrusting characters will not have 'dead' ST levels where they gain no damage.

## Change the way swing works
Swing's gone.  Instead, swing is a modifier on thrust damage based on weapon used.  +25% for reach 1+ swing, with another 25% for each of two-handed, crushing or unbalanced.

This makes weight based crushing attacks better, whereas before they were often left out and cutting damage was strictly superior except against the very rare enemies immune to it. Two-handed weapons are also far better; instead of a mere +1 for wielding a weapon with two hands, a user now gets extra muscle powered damage. At low levels of strength this IS equivalent to +1 (especially if using the , as intended. But while with RAW damage this is quickly made irrelevant, now that  extra continues even for strong characters.
Furthermore, unbalanced weapons now gain a bonus to counteract their large downside of not being able to parry.

- Alternately, some GMs may wish to have the crushing not confer a percentage bonus and have 2 handed weapons gain +50% instead.


richardwilkes:
Please make sure the IntelliJ formatter with the checked-in settings has been run on the source files you changed, or manually go in and make the files format the same way (using the formatter is a lot less tedious...).

Also, for things like switch statements, collapse cases that resolve to the same thing, i.e. instead of:

```
case 1:
    ...do some stuff...
    break;
case 2:
    ...do some stuff...
    break;
```

make it:

```
case 1:
case 2:
    ...do some stuff...
    break;
```

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
```diff
@@ -55,12 +59,14 @@
     private double         mFragmentationArmorDivisor;
     private String         mFragmentationType;
     private int            mModifierPerDie;
+    private String         mPercentBonus;
```
richardwilkes, line 62: This should be stored as an integer, not as a string, so that any other calculations that may want to use it in the future don't have to muck with string conversion.

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
```diff
@@ -29,12 +29,15 @@
...
 import com.trollworks.gcs.utility.text.Enums;
 import com.trollworks.gcs.utility.text.Numbers;
+import com.trollworks.gcs.settings.DamageProgression;
 
 import java.io.IOException;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Set;
 
+import javax.print.attribute.standard.Sides;
```
richardwilkes, line 39: It is unclear to me why you're pulling in the Sides option for the print dialog. Please remove this.

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
```diff
@@ -424,8 +479,22 @@ private void extractWeaponDamageBonus(Feature feature, Set<WeaponDamageBonus> se
         }
     }
 
-    private static Dice addDice(Dice left, Dice right) {
-        return new Dice(left.getDieCount() + right.getDieCount(), Math.max(left.getDieSides(), right.getDieSides()), left.getModifier() + right.getModifier(), left.getMultiplier() + right.getMultiplier() - 1);
+    private Dice addDice(Dice left, Dice right) {
+        
+        if(mOwner.mOwner.getDataFile().getSheetSettings().getDamageProgression() == DamageProgression.PHOENIX_D3){
```
richardwilkes, line 484: Perhaps you should only do this for weapons that are actually using d6. In GURPS, I think that is currently all of them, however, GCS itself supports any arbitrary number of die sides in its input (and as a GM, I've occasionally made use of this). Only when the side type is left out does GCS default it to d6 (i.e. 2d resolves to 2d6, but 2d4 stays at 2d4).

Of course, that would mean that it would be impossible to actually specify d6's once this option was turned on, as you'd always downgrade them to d3's...

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
Legendsmith: Yeah, Done so! 👍

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
Legendsmith: Yeah, this is a good point. I've got an idea, I could make this code agnostic and have the 'convert to d3s' option becomes "convert dice down" or something; so the default behaviour is "turn disparate sided added dice into equivalents of the higher sided die" and then with the option on it's "turn added disparate dice into equivalents of the *lower* sided die". What do you think?

richardwilkes: Some sort of generic transformation would be good, if possible. Those titles sound awful, though. ;-)

There's also the standard GURPS option for turning adds into dice already present, which this might interact weirdly with.

As an aside, have you actually tried this new method in play? I think my players would rapidly revolt due to the annoyance of having to mentally convert all of those d6's on the table into d3 values. It isn't hard, but it can be tedious, especially if you have to do it every time you roll damage. I guess they do sell specialty dice that are labeled 1-3 twice, which would fix that... although I doubt my players would willingly buy them.

.idea/codeStyles/Project.xml
```diff
@@ -27,24 +27,6 @@
...
-      <option name="PACKAGES_IMPORT_LAYOUT">
-        <value>
-          <package name="" alias="false" withSubpackages="true" />
-          <package name="java" alias="false" withSubpackages="true" />
-          <package name="javax" alias="false" withSubpackages="true" />
-          <package name="kotlin" alias="false" withSubpackages="true" />
-          <package name="" alias="true" withSubpackages="true" />
-        </value>
-      </option>
-    </JetCodeStyleSettings>
```
richardwilkes, line 47: Please don't modify any of the settings files in the `.idea` tree. I realize your IDE might have decided to do this for you automatically, so you'll need to back this and the other one out, then be careful not to commit them again in the future.

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
```diff
@@ -272,10 +295,20 @@ public String getResolvedDamage(StringBuilder toolTip) {
                         base.multiply(advantage.getLevels());
                     }
                 }
+                //Phoenix D3
+                double percent = (double) 0;
```
richardwilkes, line 299: You should use 0.0 instead of a cast here.

com.trollworks.gcs/src/com/trollworks/gcs/utility/Dice.java
```diff
@@ -238,6 +238,15 @@ public void multiply(int multiply) {
         }
     }
 
+    public void percentAdd(double percent) {
+        if (percent != 0) {
+            int average = (mSides + 1) / 2;
+            int base    = (int) Math.floor(((mCount * average) + mModifier) * percent);
+            mCount += base / average;
```
richardwilkes, line 245: You need to guard against division by 0 here. mSides should never come in as 0 or -1, but always safer to not make the assumption.

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
```diff
@@ -237,6 +248,18 @@ public void setType(String type) {
...
+    public Integer getPercentBonus() {
+        return mPercentBonus;
+    }
+
+    public void setPercentBonus(Integer percentBonus) {
+        if (mPercentBonus != percentBonus) {
+            mPercentBonus = percentBonus;
+            notifyOfChange();
+        }
+
```
richardwilkes, line 260: Please remove dead space like this at the beginning or end of functions (I've noticed it in a few other spots, too).

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
Legendsmith: Thanks, will do. I have no formal java training so thanks again for being patient with me. This change will be committed soon™.

com.trollworks.gcs/src/com/trollworks/gcs/utility/Dice.java
richardwilkes: This still doesn't guard against `average` becoming 0 and then being used in division.

richardwilkes:
Made a few more comments. I haven't gone over the changes in-depth yet. When you think you're done, let me know and I'll do a more careful pass.

com.trollworks.gcs/src/com/trollworks/gcs/utility/Dice.java
Legendsmith: I knew I forgot something! Thanks.

Legendsmith: I see now, average never should have been an int, I braindead-ly just wrote it as such.

com.trollworks.gcs/src/com/trollworks/gcs/weapon/WeaponDamage.java
Legendsmith: Well, there's actually various d3s you can get; there's dedicated d3 shapes. There's also 6 sided die that only have 1 to 3; saves the mental conversion! We've only done some tests at this point, rather than like a full campaign with this though.

Legendsmith:
I think this is ready now. If there are any other changes to be made at this point they'll be very minor.

richardwilkes:
I want some additional changes, but since I failed to actually click the button that asked for them last week, I'll just merge what you have and make them myself. Sorry about that.

PR #428, opened 2022-03-04
Add dodge to techniques

ryder-james:
This change adds dodge to the list of options for techniques.
This is my first change to GCS and I'm a little nervous about having done it right, but it works, so let me know if anything needs to change.

As far as I can tell, this change seems to be only particularly useful for the Timed Dodge technique on GF29.


Calmquist (approved):
I am not particularly familiar with the design, but it looks correct.

richardwilkes:
I've applied this change to the new code base, but am currently holding off on applying any changes at all to the Java code base.

richardwilkes:
This will be in the v5 release

Issue #518, opened 2022-09-07
v5: Technique Difficulty shows as "DX/*" in Skill Library


rinickolous:
The value displays correctly for Skills, but Techniques seem to default to showing "DX/" as the attribute portion of the difficulty value.

richardwilkes:
Hmm... I don't see this. Can you give me step-by-step details to reproduce?

rinickolous:
This seems to not be a display issue as much as an issue when creating new Techniques. They are automatically assigned difficulty value "dx/a" instead of "a". To reproduce:

1. Create a new Technique.
2. Drag into a skill library (or check file) to verify value.
3. Value will be displayed as "DX/A" instead of expected "A".

PR #520, opened 2022-09-08
fixes richardwilkes/gcs#518

ImmanuelMc:
FormatRelativeSkill doesn't handle Techniques and resorts to default

Added FormatRelativeTechniqueSkill to handle calls for Techniques. If FormatRelativeSkill can be safely changed to take string instead of Difficulty,  the skills could be consolidated, but wasn't sure if that would cause problems elsewhere.


ImmanuelMc:
Noticed a problem

PR #521, opened 2022-09-08
fixes richardwilkes/gcs#518

ImmanuelMc:
Techniques keep their relevant data differently than skills do. So I created a FormatRelativeTechniqueSkill to handle techniques.

### DIFF
--- a/model/gurps/skill.go
+++ b/model/gurps/skill.go
@@ -223,7 +223,11 @@ func (s *Skill) CellData(column int, data *CellData) {
 	case SkillRelativeLevelColumn:
 		if !s.Container() {
 			data.Type = Text
-			data.Primary = FormatRelativeSkill(s.Entity, s.Type, s.Difficulty, s.AdjustedRelativeLevel())
+			if strings.HasPrefix(s.Type, gid.Technique) {
+				data.Primary = FormatRelativeTechniqueSkill(s.Entity, s.Type, s.SkillEditData, s.AdjustedRelativeLevel())
+			} else {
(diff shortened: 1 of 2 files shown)


richardwilkes:
I tested issue #518 with your changes, but they didn't seem to address the problem. If this was to fix something else, could you describe the problem and list the steps you took that exhibited it so that I can take a look?

The fix for #518 was actually very simple: see commit bb584132eac495939b577eb1f183e1fd2694fd95.

ImmanuelMc:
I am a doofus. I was looking at that issue because it seemed straight forward, but I guess forgot the original issue while I was learning how things worked and noticed a difference in how techniques and skills display relative skill level started looking at what was going on there. 

For techniques, right now the relative skill level is just given as "+1" or "-3". My change was to make it so it says "ST+1" or "Acrobatics-3" or whatever. Dunno which is the preferred behavior.

Sorry for the confusion.

richardwilkes:
Techniques are always relative to their parent (usually a skill, but there can be non-skill techniques). Because of this, they are always displayed as just the delta (i.e. a plain +1 or -2, etc.). The canonical form used in GURPS material actually just leaves RSL blank for Techniques. You can see this in the sample character sheet for Louis d'Antares in the Basic Set (page 313 -- he only has a single technique, Feint (Force Sword), and it has no value in the Relative Level column, although it does resolve to a +2 over his normal Force Sword skill, which is what GCS shows for convenience).

ImmanuelMc:
Gotcha. Thank you for taking the time to explain.

PR #582, opened 2022-10-12
Fix usage of external PDF readers

ryder-james:
The general setting for using external PDF readers is a great addition, but it didn't respect spaces in filepaths, causing it to break somewhat spectacularly if you have those - this change wraps the result of $PATH in quotation marks so that the full filepath is respected.

### DIFF
--- a/ui/workspace/settings/pageref_mappings.go
+++ b/ui/workspace/settings/pageref_mappings.go
@@ -123,7 +123,7 @@ Would you like to create one by choosing a PDF to map to this key?`), key), pdfN
 				}
 			} else {
 				var parts []string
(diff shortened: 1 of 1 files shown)


richardwilkes:
I purposely did not do this, since just wrapping in quotes isn't correct in some situations, or requires different types of quotes, etc. Due to this, I felt it was better to require the user to know what they are doing and to properly construct the command line (i.e. fill in the quotes themselves in the field, something like `/path/to/my/viewer --page $PAGE --file "$FILE"`). Stuff like this combined with a general lack of knowledge by most GCS users of how a command line functions is the reason why I've been reluctant to add anything remotely like this for many years now.

ryder-james:
It's an entirely fair conclusion to come to. I didn't realize different situations needed different types of quotations.

ryder-james:
With that said, I'm very happy with the addition - takes since fiddling to get going but lets me continue to use my favorite PDF reader, and for that I'm grateful.

richardwilkes:
Glad it is of use. Probably should have added this a long time ago, but the potential support issues always kept me from doing so. Hopefully I won't regret adding it.

Issue #676, opened 2023-03-05
allow decimal numbers in attribute point costs


SIGSEGV111:
Hi,

When adding a new attribute to GCS for a custom house rule we need to put in a fractional numbers in the "Cost per Point" field.
The fractional numbers are the result of a 10% cost reduction that is already calculated into the point costs for the attribute.
However currently GCS does not allow this. It is not even possible to enter the "." into the field.
Could you please enable this?
Thank you!

richardwilkes:
This is already supported. You just need to create a custom set of attributes and use the Decimal type rather than the Integer type for them.

SIGSEGV111:
The decimal type only allows me to enter decimal values in the "Base Value" field, however NOT in the "Cost per Point" field. 

I would like to make the "Cost per Point" the value "2.7", bit it won't allow me to enter the "." character into the field.
Am I doing something wrong here?

![grafik](https://user-images.githubusercontent.com/8617268/222980545-763a8f84-9b46-4599-8ac5-8e54e287f8c7.png)

richardwilkes:
I misread what you were asking for. I'll take a look at your PR today.

PR #677, opened 2023-03-05
implement feature request #676

SIGSEGV111:


richardwilkes:
Thanks for the PR. I went with a slightly different bit of code, but ultimately the same result: https://github.com/richardwilkes/gcs/commit/aa1bdc42acb85281cd831aff11caa73c08c160d1

Issue #560, opened 2022-09-28
Please allow external windows again


ArchSinccubus:
I've tried to work with the new version, and I have to admit, the way it's built is *extremely* hectic and hard to go through. Creating a new Advantage is now an entirely new tab instead of external window that is stuck to the side of the screen, instead of appearing front and center. I would very much appreciate, at the very least, to be able to drag these tabs back into being external windows. The compact view of earlier versions was much more friendly to look at.

richardwilkes:
I do realize it is an extra click, but you can always hit the maximize button in the upper right of the new tab that is created to have it overlay the entire document area while you are interacting with it.

There are many reasons I moved away from modal dialogs, the most significant one being that some platforms no longer allow programs to bring windows to the foreground reliably, which means modal dialogs sometimes come up behind other windows -- yet you're locked out of interacting with any but the now-hidden window, which is very confusing and sometimes very hard to correct, since getting to that hidden window isn't always easy. This was already happening in the Java version for some people, which was very problematic.

smithkm:
I'd really appreciate being able to have multiple windows to take advantage of multiple monitors.  Dragging the one window out to span both monitors sort of works but the divider between panes doesn't always stay at the boundary.

richardwilkes:
Relevant to this issue and posted on the Discord earlier:

I've been working on allowing users to choose to open some (or even all) types of content in their own window. As I've noted before, this has been hazardous in the past, due to the way most/all of the platforms sometimes block software from bringing windows to the front. To help mitigate this, I'm making most of the windows modeless instead of modal.

Anyway, I'm looking for some folks to try it out before I do a general release to help find any bugs that may be lurking. If you're interested, please DM me.

Note: one significant drawback to this mode is that drag and drop operations between separate windows don't work at all. This is due to the underlying glfw library I use to integrate with the OS windows -- for whatever reason, that library doesn't support native drag & drop, except for file objects from the desktop. I may be able to work around this in the future by doing my own implementation (I've done this in the past on other projects), but it is a fairly significant undertaking all by itself, so I'd probably not do this any time soon.

PR #689, opened 2023-05-01
Add support for separate windows

richardwilkes:
A significant caveat to this feature: drag & drop between windows is not currently supported by the underlying windowing code (https://github.com/glfw/glfw) that GCS relies on.


richardwilkes:
This PR implements #560.

PR #806, opened 2024-01-20
Changed how minimum weapon ST is calculated

rinickolous:
According to rulings linked to [here](https://gurps.fandom.com/wiki/Striking_ST), GURPS should treat minimum ST for weapons differently depending on the type of weapon. Muscle-powered weapons, including all melee weapons as well as things like bows, should use Striking ST as a minimum. Non-muscle-powered weapons such as firearms should instead use Lifting ST.

This implementation covers the following cases correctly:
- All melee weapons, which should always use Striking ST.
- All muscle-powered ranged weapons such as bows, which use Striking ST.
- Most non-muscle-powered ranged weapons such as firearms, which use Lifting ST.
However, it covers some caes incorrectly, including:
- Some non-muscle-powered ranged weapons such as Crossbows, which use Lifting ST but cannot be differentiated from bows using the implemented model.

The implementation is less wrong than the current one which uses Striking ST in all cases, but currently fails to differentiate between bows and crossbows. I cannot think of a way to accomodate for this without checking against the skill used by the weapon. Doing so would cover all cases but would not be localization-friendly, so I would like to avoid it if possible.

### DIFF
--- a/model/gurps/weapon.go
+++ b/model/gurps/weapon.go
@@ -374,7 +374,16 @@ func (w *Weapon) SkillLevel(tooltip *xio.ByteBuffer) fxp.Int {
 
 func (w *Weapon) skillLevelBaseAdjustment(e *Entity, tooltip *xio.ByteBuffer) fxp.Int {
 	var adj fxp.Int
-	if minST := w.Strength.Resolve(w, nil).Min - e.StrikingStrength(); minST > 0 {
+	var minST fxp.Int
+	// If the weapon is ranged and does not use thrust/swing, we can infer that it is not muscle powered
+	// such as firearms. Melee weapons and bows use striking ST, while non-muscle-powered weapons use
+	// lifting ST.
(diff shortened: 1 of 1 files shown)


richardwilkes:
Out of curiosity, what ranged weapons would actually use Lifting ST instead of Striking ST? The use of that bonus is exclusive to weapons whose damage is based on ST and aren't using the "weapon ST" capability (i.e. not bows/crossbows). What else is there?

richardwilkes:
Ah... I just realized that this is actually about the Min ST field, not the Damage field (despite it being right there in your initial text... I just glossed over that).

richardwilkes:
Implemented something similar in commit ca494c0f46460b841c26f2582304e5261405852b. Note that this does explicitly check for the use of the Crossbow skill to handle the crossbow case, which seems to be a special case. I could have done this by adding yet another boolean field in the weapon data, however, it would not have been set correctly for existing entries, so decide against it.

PR #889, opened 2024-10-23
Only send deflate encoded responses when supported

fishface60:
deflate is normally well-supported across browsers but MapTool only supports gzip out of the box.

It indicates this correctly by only sending gzip in Accept-Encoding though so we can test for this and only compress if it includes deflate.

I have submitted https://github.com/RPTools/maptool/pull/5016 to add support to MapTool also so it can support compressed encoding when the next release comes out.

### DIFF
--- a/server/sheets.go
+++ b/server/sheets.go
@@ -124,7 +124,7 @@ func (s *Server) sheetHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	response := sheet.NewSheetFromEntity(entity.Entity, entity.OriginalHash != entity.CurrentHash, access.ReadOnly)
-	CompressedJSONResponse(w, http.StatusOK, response)
+	PossiblyCompressedJSONResponse(w, r, http.StatusOK, response)
 }
 
 func (s *Server) updateSheetHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


server/sheets.go
```diff
@@ -118,13 +118,28 @@ func prune(dirs []*Dir) []*Dir {
 	return dirs
 }
 
+func hasDeflate(acceptEncoding []string) bool {
+	for _, s := range acceptEncoding {
+		for _, encoding := range strings.Split(s, ", ") {
```
richardwilkes, line 123: Probably should not rely on there being a space after the comma and instead split just on the comma and then use strings.TrimSpace() to checking for deflate.

server/sheets.go
```diff
@@ -118,13 +118,28 @@ func prune(dirs []*Dir) []*Dir {
...
+}
+
 func (s *Server) sheetHandler(w http.ResponseWriter, r *http.Request) {
 	entity, access, ok := s.loadSheet(w, r)
 	if !ok {
 		return
 	}
 	response := sheet.NewSheetFromEntity(entity.Entity, entity.OriginalHash != entity.CurrentHash, access.ReadOnly)
-	CompressedJSONResponse(w, http.StatusOK, response)
+	if hasDeflate(r.Header.Values("Accept-Encoding")) {
```
richardwilkes, line 138: I think I'd prefer to have a function that does this so that the same if/else doesn't need to be repeated everywhere. Perhaps PossiblyCompressedJSONResponse()?

richardwilkes:
Generally looks good, but would like a few tweaks before accepting it. Thanks!

fishface60:
Thanks, suggestions applied and re-tested that MapTool is able to read the response and that in the webapp the response uses Content-Encoding: deflate.

richardwilkes (approved):
Thanks!

PR #990, opened 2025-09-09
Groundwork for Skill Trees

NocTempre:
Functionality currently limited to allowing default chaining with sheet setting selected.


model/gurps/skill.go
```diff
@@ -524,10 +524,11 @@ func (s *Skill) DefaultSkill() *Skill {
 	if e == nil {
 		return nil
 	}
+	requirePoints := e.SheetSettings.UseSkillTrees
```
richardwilkes, line 527: Given the default should be false for UseSkillTrees, this (and the next change) is essentially inverting the current logic, which shouldn't be the case.

model/gurps/skill.go
```diff
@@ -732,8 +733,10 @@ func (s *Skill) DecrementSkillLevel() {
 func (s *Skill) CalculateLevel(excludes map[string]bool) Level {
 	points := s.AdjustedPoints(nil)
 	if s.IsTechnique() {
-		return CalculateTechniqueLevel(EntityFromNode(s), s.Replacements, s.NameWithReplacements(),
-			s.SpecializationWithReplacements(), s.Tags, s.TechniqueDefault, s.Difficulty.Difficulty, points, true,
+		e := EntityFromNode(s)
```
richardwilkes, line 736: This may return nil, so you can't just use it below with additional safe-guards.

model/gurps/skill.go
```diff
@@ -732,8 +733,10 @@ func (s *Skill) DecrementSkillLevel() {
...
+		requirePoints := !e.SheetSettings.UseSkillTrees
```
richardwilkes, line 737: Again, the use of this seems to be inverting the current logic, which is incorrect.

model/gurps/skill.go
```diff
@@ -916,11 +921,11 @@ func (s *Skill) bestDefault(excluded *SkillDefault) *SkillDefault {
 
 func (s *Skill) calcSkillDefaultLevel(def *SkillDefault, excludes map[string]bool) fxp.Int {
 	e := EntityFromNode(s)
-	level := def.SkillLevel(e, s.Replacements, true, excludes, !s.IsTechnique())
-	if def.SkillBased() {
+	level := def.SkillLevel(e, s.Replacements, !e.SheetSettings.UseSkillTrees, excludes, !s.IsTechnique())
```
richardwilkes, line 924: You can't assume e has a non-nil value here

model/gurps/skill.go
```diff
@@ -916,11 +921,11 @@ func (s *Skill) bestDefault(excluded *SkillDefault) *SkillDefault {
...
+	if def.SkillBased() && e != nil && e.SheetSettings.UseSkillTrees {
```
richardwilkes, line 925: The old code did not skip this computation if there was no entity, so this seems wrong again.

model/gurps/skill.go
```diff
@@ -973,6 +980,9 @@ func (s *Skill) TechniqueSatisfied(tooltip *xbytes.InsertBuffer, prefix string)
 		return true
 	}
 	e := EntityFromNode(s)
+	if e.SheetSettings.UseSkillTrees {
+		return true
```
richardwilkes, line 984: Do techniques even exist in the skill tree world? I read it when it was released, but haven't looked since, so may just be not remembering them being there.

ux/sheet_settings_dockable.go
```diff
@@ -52,6 +52,7 @@ type sheetSettingsDockable struct {
 	useHalfStatDefaults                *unison.CheckBox
 	showLiftingSTDamage                *unison.CheckBox
 	showIQBasedDamage                  *unison.CheckBox
+	UseSkillTrees                      *unison.CheckBox
```
richardwilkes, line 55: Unless there is a good reason to make this public, don't.

ux/sheet_settings_dockable.go
```diff
@@ -221,6 +222,11 @@ func (d *sheetSettingsDockable) createOptions(content *unison.Panel) {
 			d.settings().ShowIQBasedDamage = d.showIQBasedDamage.State == check.On
 			d.syncSheet(false)
 		})
+	d.UseSkillTrees = d.addCheckBoxWithLink(panel, i18n.Text("Use Skill Trees"), "PU10:1+",
```
richardwilkes, line 225: That link should be to a specific page, not something with a trailing + -- that format is just used for display, not for navigation.

richardwilkes (changes requested):

build.sh
```diff
@@ -56,7 +56,7 @@ for arg in "$@"; do
 		BUILD_GEN=1
 		SOMETHING=1
 		;;
-	--custom | -c)
+	--win | -w)
```
richardwilkes, line 59: Still not grasping why this is here. I build for Windows all the time and this is not needed. Specifically, the -s option disables symbols in the output and the -w option disables DWARF generation. Both of those should only be on the command line when building for a release. By setting them, you're making debugging harder for yourself.

If you're doing this because of some oddity in your particular setup (I believe you were the one having tons of difficulty compiling on Windows for some reason), it's unclear why this would help. It shouldn't have any effect with regards to that... and if it does, it just indicates there is something fundamentally wrong in your setup.

NocTempre: Hard to say, I'm very new to GO.  I assumed the -w was a windows flag.  I tried just -s and just -w and either and both was ok.  Neither and it can't launch as an .exe for me.

richardwilkes:
I'm finally back to working on GCS-related things. Unsurprisingly, my recent updates have some conflicts with this PR. I'm likely going to be doing some work to support Mission: X in the future, which this stuff probably overlaps with given the way Doug's system is shaking out.

richardwilkes:
I'm going to close this since I haven't seen any updates or responses. Feel free to reopen if you do come back to it.

Issue #1023, opened 2026-02-16
External PDF viewer can't highlight search term


justanothercell:
Unlike the internal PDF viewer, the external viewer has no way of highlighting the searched phrase.
The file and page are substituted in the command like args with `$FILE` and `$PAGE` respectively and the queried word should be passed similarly, e.g. via `$WORD`.

Example usage: `firefox --new-tab "file://$FILE#page=$PAGE&search=$WORD"`

PR #1024, opened 2026-02-16
Pass highlighted phrase to external PDF viewer via `$WORD`

justanothercell:
Implements #1023.

Update signature of `openExternalPDf(filePath string, pageNum int)` to `openExternalPDF(filePath string, highlight string, pageNum int)`, aswell as the two call sites and substitute `$WORD` in the external pdf command, similar to the existing `$FILE` and`$PAGE`.

In [navigator.go](https://github.com/richardwilkes/gcs/blob/master/ux/navigator.go#L1248) `openExternalPDF(absPath, 1)` is used with 1 as the default initial page, despite `initialPage` being an available value. In any case there is no `highlight` value available, so the signature was updated to use an empty string as the highlight: `openExternalPDF(absPath, "", 1)`.

I don't think it is possible to directly suggest a change to the wiki, but the `$WORD` should be mentioned in the [external pdf settings](https://github.com/richardwilkes/gcs/wiki/General%20Settings#external-pdf-settings) entry. (I could e.g. create a git patch or paste the markdown for a proposed edit somewhere here)

### DIFF
--- a/ux/general_settings_dockable.go
+++ b/ux/general_settings_dockable.go
@@ -410,10 +410,10 @@ func (d *generalSettingsDockable) createExternalPDFCmdLineField(content *unison.
 	}
 	d.externalPDFCmdlineField.Tooltip = newWrappedTooltip(i18n.Text(`The internal PDF viewer will be used if the External PDF Viewer field is empty.
 Use $FILE where the full path to the PDF should be placed.
-Use $WORD where the highlighted string should be placed.
+Use $TEXT where the highlighted string should be placed.
 Use $PAGE where the page number should be placed.
 
-In most cases, you'll want to surround $FILE and $WORD with quotes.
+In most cases, you'll want to surround $FILE and $TEXT with quotes.
 Note that this might still fail, e.g. if the variable itself contains quotes.`))
 	content.AddChild(d.externalPDFCmdlineField)
 }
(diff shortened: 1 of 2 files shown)


ux/navigator.go
```diff
@@ -1245,7 +1245,7 @@ func OpenFile(filePath string, initialPage int) (dockable unison.Dockable, wasOp
 		return nil, false
 	}
 	if fi.IsPDF && strings.TrimSpace(gurps.GlobalSettings().General.ExternalPDFCmdLine) != "" {
-		openExternalPDF(absPath, 1)
+		openExternalPDF(absPath, "", 1)
```
richardwilkes: Let's fix this to use `max(initialPage,1)` here as well. Likely just an oversight from before `initialPage` was passed into this function. Note that 0 is passed from other parts of the code in many cases, so ensuring the page is at least 1 is important for some PDF viewers.

richardwilkes (commented):
You should also update the string on lines 411-415 of ux/general_settings_dockable.go to reflect the new $WORD variable.

Like $FILE, you should note that it likely should be surrounded with quotes. This will still fail under some circumstances (for example, when the substitution itself contains a quote), but there isn't a reasonable way to prevent that which will work on all supported platforms and shells.

richardwilkes (commented):
One more minor change and then I'll merge it. Thanks!

PR #1028, opened 2026-03-10
Save GCS version to data files

### LINKED ISSUE #1027: Save GCS version into data files
To help with diagnosing data loading issues, it is sometimes useful to know which version of GCS actually wrote the file. To that end, we should add the GCS release version into each saved file's data.

utafrali:
Fixes #1027

Stores the GCS version in the saved Campaign, Entity, and Loot files. Useful for tracking which version created each file. Just populates gcs_version from xos.AppVersion on save.


utafrali:
gonna close this for now, doesn't seem like it's needed anymore

PR #1034, opened 2026-04-10
Changing metric length conversion to 25mm per inch

kirkeugle:
Changed length_unit.go 
- to use 2.5cm per inch to convert to metric
- to provide LengthUnit.FromInches(int) converter
- separated conversion from formating

Changed calculator.go to use correct conversions
- also round to next whole km for hiking

### DIFF
--- a/model/fxp/length_unit.go
+++ b/model/fxp/length_unit.go
@@ -33,20 +33,34 @@ func (enum LengthUnit) Format(length Length) string {
 			buffer.WriteByte('"')
 		}
 		return buffer.String()
-	case Inch:
-		return inches.String() + " " + enum.Key()
-	case Feet:
-		return inches.Div(Twelve).Comma() + " " + enum.Key()
-	case Yard, Meter:
-		return inches.Div(ThirtySix).Comma() + " " + enum.Key()
-	case Mile:
-		return inches.Div(MileInInches).Comma() + " " + enum.Key()
-	case Centimeter:
-		return inches.Div(ThirtySix).Mul(Hundred).Comma() + " " + enum.Key()
-	case Kilometer:
-		return inches.Div(ThirtySixThousand).Comma() + " " + enum.Key()
+
 	default:
-		return FeetAndInches.Format(length)
+		return enum.FromInches(inches).Comma() + " " + enum.String()
+	}
+}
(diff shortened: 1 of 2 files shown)


richardwilkes:
Looks good. As I think we already discussed, I originally chose to use the 1 yard -> 1 meter conversion from B9. Since I didn't want the distance to change based on the units, I applied that across the board, unlike the chart that shows different conversion rates depending on unit. Using 1 inch -> 25mm appears to be a better choice, though.

I wonder if a similar change from 1 lb -> 0.5 kg to 1 lb -> 0.45 kg might also be warranted? Do you have an opinion?

cognato-net:
> I wonder if a similar change from 1 lb -> 0.5 kg to 1 lb -> 0.45 kg might also be warranted? Do you have an opinion?

I'm not the original poster, but since you asked: I play almost exclusively in metric, and I would prefer not to change GCS's general conversion rates unless it is added as an optional setting.

Updating the character profile/description section makes sense to me. It is quite quirky to see newly created characters listed as, say, `1.8611 m` tall, and while I personally didn't notice the specific height issue kirkeugle described, I'm glad to see it fixed. Along the same lines, checking and improving the weight displayed in the profile/description section would be very welcome too.

But I would be much more cautious about changing item/trait weights, distances, or other values that affect the main mechanics. The simplified conversions (`1 lb = 0.5 kg`, `1 yd = 1 m`, `1 mi = 1.5 km` ) have been standard in Brazilian GURPS (and, I imagine, elsewhere internationally) for decades. As far as I know, Steve Jackson introduced them when GURPS was first translated, and they have been firmly established since 3e.

Changing those conversions globally would create serious compatibility problems. The published Brazilian GURPS books list item weights in kg using the `1 lb = 0.5 kg` approximation. So even though `1 lb = 0.45 kg` is more accurate in real-world terms, it would complicate things if an item that my entire table treats as weighing `0.5 kg` appeared in GCS as `0.45 kg` instead.

I can definitely see the benefit of more precise conversions when the whole table wants that, especially if everyone is using GCS. But when not everyone is, particularly the GM, changing the conversion rates globally could make GCS less compatible with GURPS as written and as actually played in metric-language editions.

PR #1035, opened 2026-04-10
Calculator: Added hours of hiking per day and hiking distance

kirkeugle:
This commit addresses two things
1. It adjust hiking distance according to the clarification in HT55
  - Introducing a drop down menu providing hiking hours per in 4h increment + custom
  - Default hiking duration is set to a more typical value of 8h
2. Introduced a field for total distance and a calculation how long it would take in days with current hiking parameters (I hope someone finds that helpful)

### DIFF
--- a/model/fxp/int.go
+++ b/model/fxp/int.go
@@ -21,6 +21,7 @@ import (
 var (
 	Min                 = fixed64.Minimum[DP]()
 	NegPointEight       = FromStringForced("-0.8")
+	Zero                = FromInteger(0)
 	OneHundredth        = FromStringForced("0.01")
 	Twentieth           = FromStringForced("0.05")
 	PointZeroSix        = FromStringForced("0.06")
--- a/ux/calculator.go
+++ b/ux/calculator.go
@@ -66,6 +66,14 @@ var (
 		{Name: i18n.Text("Snow"), Modifier: fxp.Half, IsSnow: true},
 		{Name: i18n.Text("Snow, Heavy"), Modifier: fxp.Quarter, IsSnow: true},
 	}
+
+	hikingIntensity = []hikingIntensityHours{
+		{Name: i18n.Text("Forced March"), HoursHiking: fxp.Sixteen},
+		{Name: i18n.Text("Long March"), HoursHiking: fxp.Twelve},
+		{Name: i18n.Text("Normal"), HoursHiking: fxp.Eight, Default: true},
+		{Name: i18n.Text("Foraging"), HoursHiking: fxp.Four},
+		{Name: i18n.Text("Custom"), IsCustom: true},
+	}
 )
 
 type terrainModifier struct {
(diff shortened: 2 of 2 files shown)


model/fxp/int.go
```diff
@@ -21,6 +21,7 @@ import (
 var (
 	Min                 = fixed64.Minimum[DP]()
 	NegPointEight       = FromStringForced("-0.8")
+	Zero                = FromInteger(0)
```
richardwilkes: This constant is unnecessary, as the untyped constant '0' is the same. Please remove.

richardwilkes (changes requested):
Could I get you to fix the merge conflicts before I review this further? Thanks!

model/fxp/int.go
kirkeugle: Done

ux/calculator.go
```diff
@@ -222,7 +254,7 @@ func (c *Calculator) createContent() {
 }
 
 func (c *Calculator) addJumpingSection() {
-	c.content.AddChild(c.createHeader(i18n.Text("Jumping"), "BX352", "Jumping", 0))
+	c.content.AddChild(c.createHeader(i18n.Text("Jumping"), []linkSpec{{"BX352", "Jumping"}}, 0))
```
richardwilkes: I generally don't like to rely on the position of fields within a struct when assigning data, since I frequently rearrange the fields as more things are added, often to improve memory locality. Instead, please be explicit during initialization, e.g.
```
[]linkSpec{
    {
        pageRef: "BX352",
        highlight: "Jumping",
    },
}
```
I know it is a lot more verbose, but eliminates errors later on should something move around. I won't note this on the other cases of it, but would like those changed as well.

ux/calculator.go
```diff
@@ -518,9 +619,9 @@ func (c *Calculator) createResultLabel() *unison.Label {
 	return label
 }
 
-func (c *Calculator) createHeader(text, linkRef, linkHighlight string, topMargin float32) *unison.Panel {
+func (c *Calculator) createHeader(text string, linkSpecs []linkSpec, topMargin float32) *unison.Panel {
 	wrapper := unison.NewPanel()
-	wrapper.SetLayout(&unison.FlexLayout{Columns: 3})
+	wrapper.SetLayout(&unison.FlexLayout{Columns: 2 + 2*len(linkSpecs) - 1})
```
richardwilkes: Why not `1 + 2*len(linkSpecs)` instead?

ux/calculator.go
```diff
@@ -941,8 +1059,20 @@ func (c *Calculator) updateHikingResult() {
...
+
+	if timeInDays == 1 {
+		c.hikingTimeLabel.SetTitle(i18n.Text("1 day"))
+	} else {
+		c.hikingTimeLabel.SetTitle(fmt.Sprintf(i18n.Text("%s days"), timeInDays))
+	}
+
 	c.hikingResult.MarkForLayoutRecursivelyUpward()
+	c.hikingTimeLabel.MarkForLayoutRecursivelyUpward()
+	c.hikingDistanceLabel.MarkForLayoutRecursivelyUpward()
```
richardwilkes: I suspect only one of these three lines needs to actually be here. The other two may need a `.NeedsLayout = true` call, but there shouldn't be a need to run up the chain for all three.

richardwilkes (changes requested):
Some minor changes I'd like to see before I merge.

Overall, looks good. Thanks.

ux/calculator.go
kirkeugle: Done. Totally agree.

ux/calculator.go
kirkeugle: Done

ux/calculator.go
kirkeugle: turns out it is sufficient to mark hikingTimeLabel

kirkeugle:
Also fixed a formating issue

richardwilkes (approved):

Issue #1038, opened 2026-04-30
Add support for Optional Specialties


rinickolous:
## Rule Definition

Optional specialties are described in Basic Set pp. 169-170 and are functionally distinct from required specialties.

Optional specialties represent sub-fields of an IQ-based skill which a character may specialize in but are not required to. They have the following effects on a skill:
- Including an optional specialty in a skill decreases the difficulty level for said skill by one level (e.g. Hard -> Average). As such, they are only compatible with skills of an original difficulty of Average or harder.
- The general skill defaults to the optionally specialized version at -2.
- Skills which default to a general skill default to the optionally specialized skill at their specified default modifier, plus an additional -2.

Importantly, a skill can have _both_ a required and optional specialty. Examples in Basic Set include:
- Biology, which must be specialized by planet type, may also be specialized by class of animal
- Artist, which must be specialized by art form, may also be specialized by medium or technique.

## The Problem 

GCS currently does not fully support some of the above features. It can support some of these features partially, but said implementation on the data side of things is clunky at best.
- Multiple replacement strings _can_ be placed within a skill, but these are not recognized as distinct specialties by the Skill Default field, which supports only an exact case-insensitive match.
- A skill defaulting to an optionally specialized version of itself at -2 is achievable by adding a Skill Default entry to the skill with the specialization left blank (for skills with only an optional specialty) but the inexact match for skills with both a required and optional specialty means that this is not possible for such skills. (This is, again, caused by the exact case-insensitive match of the specialization field for Skill Defaults).
- The difficulty level of a skill _can_ be manually adjusted if it includes an optional specialty, but this requires duplication of any skills which fall under this rule, which at least somewhat negatively affects maintainability.

### 1. Optional Specialty Field(s)

Introduce a separate "Optional specialization" field for skills. There are a few ways to do this, but this part of the implementation should meet the following requirements:
- The skill must be able to signal that it supports or does not support an optional specialty (for example, when applying a template).
- The skill must be able to remain without an optional specialty if none is specified.

I believe the following implementations would both work, and have their own strengths and weaknesses:
- 1 string field, supporting replacement functionality, synchronized. If this string is blank in the library item (does not include an `@Optional specialty@` string on the like), the skill does not support optional specialties. If the string is specified but is effectively blank due to the corresponding replacements being blank, the skill is treated as not having an optional specialty defined.
**Pros:** This requires only one additional field.
[...]

PR #1039, opened 2026-04-30
Feature: Support for Optional Specialties

rinickolous:
After writing the Issue for this (#1038 ) I decided I've probably thought it through enough to just implement this myself.

## Rule Definition

Optional specialties are described in Basic Set pp. 169-170 and are functionally distinct from required specialties.

Optional specialties represent sub-fields of an IQ-based skill which a character may specialize in but are not required to. They have the following effects on a skill:
- Including an optional specialty in a skill decreases the difficulty level for said skill by one level (e.g. Hard -> Average). As such, they are only compatible with skills of an original difficulty of Average or harder.
- The general skill defaults to the optionally specialized version at -2.
- Skills which default to a general skill default to the optionally specialized skill at their specified default modifier, plus an additional -2.

Importantly, a skill can have _both_ a required and optional specialty. Examples in Basic Set include:
- Biology, which must be specialized by planet type, may also be specialized by class of animal
- Artist, which must be specialized by art form, may also be specialized by medium or technique.

## The Problem 

GCS currently does not fully support some of the above features. It can support some of these features partially, but said implementation on the data side of things is clunky at best.
- Multiple replacement strings _can_ be placed within a skill, but these are not recognized as distinct specialties by the Skill Default field, which supports only an exact case-insensitive match.
- A skill defaulting to an optionally specialized version of itself at -2 is achievable by adding a Skill Default entry to the skill with the specialization left blank (for skills with only an optional specialty) but the inexact match for skills with both a required and optional specialty means that this is not possible for such skills. (This is, again, caused by the exact case-insensitive match of the specialization field for Skill Defaults).
- The difficulty level of a skill _can_ be manually adjusted if it includes an optional specialty, but this requires duplication of any skills which fall under this rule, which at least somewhat negatively affects maintainability.

### `Skill` struct (`model/gurps/skill.go`)

- Added `OptionalSpecialization` string field with full nameable-replacement support (displayed after any required specialization, separated by a comma, e.g. `Artist (Painting, Oil)`).
- Added `AdjustedDifficulty()` returning the effective `AttributeDifficulty` — one step lower than the declared difficulty when an optional specialization is present, capped at Easy.
- `CellData()` and level display in the editor now use the adjusted difficulty rather than the raw declared difficulty.
- `resolveToSpecificDefaults()` now implements the two automatic defaulting rules:
  - **Rule 1** — if skill `Foo` has no optional specialization and `Foo (Bar)` (with optional spec `Bar`) exists on the character with points, `Foo` automatically defaults to `Foo (Bar)` at −2, without requiring an explicit Skill Default entry.
  - **Rule 2** — if skill `A` explicitly defaults to `C` at penalty `−X`, and `C (E)` (with optional spec `E`) exists on the character, `A` also defaults to `C (E)` at `−(X+2)`.
- Fixed `bestDefault()` incorrectly bailing out when a skill has no explicit defaults, which prevented Rule 1 from firing.
- Fixed `inDefaultChain()` passing a nil excludes set to the skill lookup, causing false-positive cycle detection that suppressed all same-name optional-spec defaults.

### `SkillDefault` struct (`model/gurps/skill_default.go`)

- `Name` and `Specialization` fields changed from plain `string` to `criteria.Text`, enabling all string-comparison modes (is, is not, contains, starts with, ends with, …) rather than only exact case-insensitive matching.
- Added `migrateStringToCriteriaText()` for backwards-compatible JSON loading: existing files that store these fields as plain strings are automatically migrated to `{compare: "is", qualifier: "…"}` on load.
- `best()` and `bestFast()` now call `Entity.SkillMatching()` with the full `criteria.Text` objects, so user-configured comparison modes are honoured when resolving default levels.

**Possible Issue:** The `TechniqueDefault` object uses the `SkillDefault` structure but the UI does not expose the `Compare` value for editing, the value is always `criteria.IsText`. This is intentional, as I don't think the changes made to these fields in `SkillDefault` are necessary to carry over to the Technique data, but it could potentially cause issues if the value is manually edited. This may be worth revising.

### `Entity` (`model/gurps/entity.go`)

- Added `SkillMatching(nameCriteria, specializationCriteria criteria.Text, replacements, …)` — a criteria-aware counterpart to `SkillNamed` that applies the `Compare` field rather than always using case-insensitive equality.
- Added `BestSkillMatching(…)` returning the highest-level skill among those matching.
- `SkillBonusFor()` signature updated to accept and match on `optionalSpecialization`.
- Internal skill-level resolution exclusion keys now incorporate the optional specialization to avoid false cache hits between skills that share a name and required specialization but differ in optional specialization.

### Skill editor (`ux/skill_editor.go`)

- Added **Optional Specialization** input field for non-container, non-technique skills.
- Added an **Adjusted Difficulty** read-only display next to the difficulty selector; it updates live as the optional specialization field is filled in or cleared.
- Level preview now computes using the adjusted difficulty.

### Defaults panel (`ux/defaults_panel.go`)

- `Name` and `Specialization` entries replaced with full criteria UI panels (`addNameCriteriaPanel` / `addSpecializationCriteriaPanel`), exposing the comparison-mode selector alongside the qualifier field.
- Name and specialization rows are now hidden when the default type is not skill-based, and shown when skill-based is selected.
- Modifier field now displays with a side label instead of placeholder text.

### Other call-site fixes

- `model/gurps/spell.go` — updated calls to `BestSkillNamed`, `SkillNamed`, and `AdjustedPointsForNonContainerSkillOrTechnique` to match revised signatures.
- `model/gurps/natural_attacks.go` — weapon `Defaults` entries migrated from bare string fields to `criteria.Text` via a local `skillName()` helper.

### DIFF
--- a/model/gurps/entity.go
+++ b/model/gurps/entity.go
@@ -1062,7 +1062,7 @@ func (e *Entity) SkillNamed(name, specialization string, requirePoints bool, exc
 // BestSkillMatching returns the highest-level skill whose name and specialization match the given criteria.
 func (e *Entity) BestSkillMatching(nameCriteria, specializationCriteria criteria.Text, replacements map[string]string, requirePoints bool, excludes map[string]bool) *Skill {
 	var best *Skill
-	level := fxp.Min
+	level := fxp.FromInteger(0)
 	for _, sk := range e.SkillMatching(nameCriteria, specializationCriteria, replacements, requirePoints, excludes) {
 		skillLevel := sk.CalculateLevel(excludes).Level
 		if best == nil || level < skillLevel {
--- a/model/gurps/entity.go
+++ b/model/gurps/entity.go
@@ -1081,9 +1081,7 @@ func (e *Entity) SkillMatching(nameCriteria, specializationCriteria criteria.Tex
 		if !excludes[sk.String()] {
 			if !requirePoints || sk.IsTechnique() || sk.AdjustedPoints(nil) > 0 {
 				if nameCriteria.Matches(replacements, sk.NameWithReplacements()) &&
-					(specializationCriteria.Matches(replacements, sk.SpecializationWithReplacements()) ||
-						(sk.SpecializationWithReplacements() == "" &&
-							specializationCriteria.Matches(replacements, sk.OptionalSpecializationWithReplacements()))) {
+					(specializationCriteria.Matches(replacements, sk.SpecializationWithReplacements()) || specializationCriteria.Matches(replacements, sk.OptionalSpecializationWithReplacements())) {
 					list = append(list, sk)
 				}
 			}
--- a/model/gurps/entity.go
+++ b/model/gurps/entity.go
@@ -511,6 +511,7 @@ func (e *Entity) processPrereqs() {
 					penalty.NameCriteria.Qualifier = s.NameWithReplacements()
 					penalty.SpecializationCriteria.Compare = criteria.IsText
 					penalty.SpecializationCriteria.Qualifier = s.SpecializationWithReplacements()
+					penalty.OptionalSpecializationCriteria.Qualifier = s.OptionalSpecializationWithReplacements()
 					if s.TechLevel != nil && *s.TechLevel != "" {
 						penalty.Amount = -fxp.Ten
 					} else {
@@ -830,7 +831,7 @@ func (e *Entity) AddDRBonusesFor(locationID string, tooltip *xbytes.InsertBuffer
 }
 
 // SkillBonusFor returns the total bonus for the matching skill bonuses.
-func (e *Entity) SkillBonusFor(name, specialization string, tags []string, tooltip *xbytes.InsertBuffer) fxp.Int {
+func (e *Entity) SkillBonusFor(name, specialization, optionalSpecialization string, tags []string, tooltip *xbytes.InsertBuffer) fxp.Int {
 	var total fxp.Int
 	for _, bonus := range e.features.skillBonuses {
 		if bonus.SelectionType == skillsel.Name {
@@ -840,6 +841,7 @@ func (e *Entity) SkillBonusFor(name, specialization string, tags []string, toolt
 			}
 			if bonus.NameCriteria.Matches(replacements, name) &&
 				bonus.SpecializationCriteria.Matches(replacements, specialization) &&
+				bonus.OptionalSpecializationCriteria.Matches(replacements, optionalSpecialization) &&
 				bonus.TagsCriteria.MatchesList(replacements, tags...) {
 				total += bonus.AdjustedAmount()
 				bonus.AddToTooltip(tooltip)
@@ -850,7 +852,7 @@ func (e *Entity) SkillBonusFor(name, specialization string, tags []string, toolt
 }
 
 // SkillPointBonusFor returns the total point bonus for the matching skill point bonuses.
-func (e *Entity) SkillPointBonusFor(name, specialization string, tags []string, tooltip *xbytes.InsertBuffer) fxp.Int {
+func (e *Entity) SkillPointBonusFor(name, specialization, optionalSpecialization string, tags []string, tooltip *xbytes.InsertBuffer) fxp.Int {
 	var total fxp.Int
 	for _, bonus := range e.features.skillPointBonuses {
 		var replacements map[string]string
@@ -859,6 +861,7 @@ func (e *Entity) SkillPointBonusFor(name, specialization string, tags []string, 
 		}
 		if bonus.NameCriteria.Matches(replacements, name) &&
 			bonus.SpecializationCriteria.Matches(replacements, specialization) &&
+			bonus.OptionalSpecializationCriteria.Matches(replacements, optionalSpecialization) &&
 			bonus.TagsCriteria.MatchesList(replacements, tags...) {
 			total += bonus.AdjustedAmount()
 			bonus.AddToTooltip(tooltip)
@@ -1038,7 +1041,7 @@ func (e *Entity) BestSkillNamed(name, specialization string, requirePoints bool,
 	return best
 }
 
-// SkillNamed returns a list of skills that match.
+// SkillNamed returns a list of skills that match by exact (case-insensitive) name and specialization.
 func (e *Entity) SkillNamed(name, specialization string, requirePoints bool, excludes map[string]bool) []*Skill {
 	var list []*Skill
 	Traverse(func(sk *Skill) bool {
@@ -1048,6 +1051,37 @@ func (e *Entity) SkillNamed(name, specialization string, requirePoints bool, exc
 					if specialization == "" || strings.EqualFold(sk.SpecializationWithReplacements(), specialization) {
 						list = append(list, sk)
 					}
+				}
+			}
+		}
+		return false
+	}, false, true, e.Skills...)
+	return list
+}
+
+// BestSkillMatching returns the highest-level skill whose name and specialization match the given criteria.
+func (e *Entity) BestSkillMatching(nameCriteria, specializationCriteria criteria.Text, replacements map[string]string, requirePoints bool, excludes map[string]bool) *Skill {
+	var best *Skill
+	level := fxp.Min
+	for _, sk := range e.SkillMatching(nameCriteria, specializationCriteria, replacements, requirePoints, excludes) {
+		skillLevel := sk.CalculateLevel(excludes).Level
+		if best == nil || level < skillLevel {
+			best = sk
+			level = skillLevel
+		}
+	}
+	return best
+}
+
+// SkillMatching returns a list of skills whose name and specialization match the given criteria.
+func (e *Entity) SkillMatching(nameCriteria, specializationCriteria criteria.Text, replacements map[string]string, requirePoints bool, excludes map[string]bool) []*Skill {
+	var list []*Skill
+	Traverse(func(sk *Skill) bool {
+		if !excludes[sk.String()] {
+			if !requirePoints || sk.IsTechnique() || sk.AdjustedPoints(nil) > 0 {
+				if nameCriteria.Matches(replacements, sk.NameWithReplacements()) &&
+					specializationCriteria.Matches(replacements, sk.SpecializationWithReplacements()) {
+					list = append(list, sk)
 				}
 			}
 		}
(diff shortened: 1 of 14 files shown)


rinickolous:
I haven't done this in this PR, but it may also be worth updating the data version for all files, since I think there are at least some breaking changes between the "original" version 5 and the "current" version 5.

model/gurps/entity.go
```diff
@@ -1056,6 +1059,37 @@ func (e *Entity) SkillNamed(name, specialization string, requirePoints bool, exc
 	return list
 }
 
+// BestSkillMatching returns the highest-level skill whose name and specialization match the given criteria.
+func (e *Entity) BestSkillMatching(nameCriteria, specializationCriteria criteria.Text, replacements map[string]string, requirePoints bool, excludes map[string]bool) *Skill {
+	var best *Skill
+	level := fxp.Min
```
richardwilkes: It would likely be better to set this to an initial level of 0 or maybe fxp.Two, since I believe a skill of 3 is the lowest you can roll against.

model/gurps/skill.go
```diff
@@ -255,7 +256,7 @@ func (s *Skill) IsTechnique() bool {
 func (s *Skill) Clone(from LibraryFile, owner DataOwner, parent *Skill, preserveID bool) *Skill {
 	var other *Skill
 	if s.IsTechnique() {
-		other = NewTechnique(owner, parent, s.TechniqueDefault.Name)
+		other = NewTechnique(owner, parent, s.TechniqueDefault.Name.Qualifier)
```
richardwilkes, line 265: This looks suspicious. Shouldn't it need other bits too, like the Compare value?

model/gurps/skill.go
```diff
@@ -689,7 +713,7 @@ func (s *Skill) IncrementSkillLevel() {
 	if !s.Container() {
 		basePoints := s.Points.Floor() + fxp.One
 		maxPoints := basePoints
-		if s.Difficulty.Difficulty == difficulty.Wildcard {
+		if s.AdjustedDifficulty().Difficulty == difficulty.Wildcard {
```
richardwilkes: Since wildcards don't adjust, this call to AdjustedDifficulty() isn't needed.

model/gurps/skill.go
```diff
@@ -709,7 +733,7 @@ func (s *Skill) DecrementSkillLevel() {
 	if !s.Container() && s.Points > 0 {
 		basePoints := s.Points.Floor()
 		minPoints := basePoints
-		if s.Difficulty.Difficulty == difficulty.Wildcard {
+		if s.AdjustedDifficulty().Difficulty == difficulty.Wildcard {
```
richardwilkes: Since wildcards don't adjust, this call to AdjustedDifficulty() isn't needed.

model/gurps/skill.go
```diff
@@ -925,28 +949,26 @@ func (s *Skill) calcSkillDefaultLevel(def *SkillDefault, excludes map[string]boo
...
-	for _, one := range e.SkillNamed(def.NameWithReplacements(s.Replacements),
-		def.SpecializationWithReplacements(s.Replacements), true, nil) {
+	for _, one := range e.SkillMatching(def.Name, def.Specialization, s.Replacements, true, excludes) {
 		if one == s {
 			return true
 		}
 		if !lookedAt[one] {
 			lookedAt[one] = true
-			if s.inDefaultChain(one.DefaultedFrom, lookedAt) {
+			if s.inDefaultChain(one.DefaultedFrom, lookedAt, nil) {
```
richardwilkes: Why isn't excludes being passed here?

model/gurps/skill_default.go
```diff
@@ -235,9 +283,14 @@ func (s *SkillDefault) finalLevel(level fxp.Int) fxp.Int {
 // "source" data, i.e. not expected to be modified by the user after copying from a library.
 func (s *SkillDefault) Hash(h hash.Hash) {
 	xhash.StringWithLen(h, s.DefaultType)
-	xhash.StringWithLen(h, s.Name)
-	xhash.StringWithLen(h, s.Specialization)
 	xhash.Num64(h, s.Modifier)
+	if !s.Name.IsZero() {
```
richardwilkes: These should always just call the Hash function on the criteria.Text object. Omitting them when they are zero can have unintended side effects.

ux/skill_editor.go
```diff
@@ -139,7 +133,22 @@ func initSkillEditor(e *editor[*gurps.Skill, *gurps.SkillEditData], content *uni
...
-			addDifficultyLabelAndFields(content, entity, &e.editorData.Difficulty)
+			diffWrapper := addFlowWrapper(content, i18n.Text("Difficulty"), 5)
+			addAttributeChoicePopup(diffWrapper, entity, "", &e.editorData.Difficulty.Attribute, gurps.TenFlag)
+			diffWrapper.AddChild(NewFieldTrailingLabel("/", false))
+			addPopup(diffWrapper, difficulty.Levels, &e.editorData.Difficulty.Difficulty)
+			diffWrapper.AddChild(NewFieldInteriorLeadingLabel(i18n.Text("Adjusted Difficulty"), false))
+			adjustedDiffField := NewNonEditableField(func(field *NonEditableField) {
+				localOptSpec := nameable.Apply(e.editorData.OptionalSpecialization, e.target.Replacements)
+				diff := e.editorData.Difficulty
+				if localOptSpec != "" && diff.Difficulty > difficulty.Easy {
```
richardwilkes: Shouldn't this be omitting wildcards from the reduction?

ux/skill_editor.go
```diff
@@ -154,16 +163,21 @@ func initSkillEditor(e *editor[*gurps.Skill, *gurps.SkillEditData], content *uni
...
 				var level gurps.Level
 				if e.target.IsTechnique() {
 					level = gurps.CalculateTechniqueLevel(entity, e.target.NameableReplacements(), localName, localSpec,
 						e.editorData.Tags, e.editorData.TechniqueDefault, e.editorData.Difficulty.Difficulty, points,
 						true, e.editorData.TechniqueLimitModifier, nil)
 				} else {
-					level = gurps.CalculateSkillLevel(entity, localName, localSpec, e.editorData.Tags,
-						e.editorData.DefaultedFrom, e.editorData.Difficulty, points,
+					adjDiff := e.editorData.Difficulty
+					if localOptionalSpec != "" && adjDiff.Difficulty > difficulty.Easy {
```
richardwilkes: Shouldn't this also be omitting wildcards from the reduction?

richardwilkes (commented):
Hmm... in over 30 years of playing GURPS, I don't think I've ever had a single player that used the "optional" specialties.

The presentation is now a bit confusing, as the specialization field has always been treated as optional, but now there is a separate optional specialization field, implying that the first one is not optional at all. Should probably rename the first one to "Required Specialization", as I think that's how they are referred to in the rules (haven't verified that, though).

model/gurps/skill.go
rinickolous: This is actually something I would like some feedback about. Due to the changes I made to `SkillDefault`, `s.TechniqueDefault.Name` splits into `.Qualifier` and `.Compare` by virtue of it being a `SkillDefault` struct, but it does not make sense to have a `.Compare` value other than `compare.IsText` for Techniques in the same way that it does for Skill Defaults in a list. This basically means that the struct in the data will have a `.compare` value which is not editable in GCS itself. It may be wiser to use a different struct (which just has the same structure as the previous `SkillDefault` struct with the name and specialization as normal strings rather than compare structs), but I want to run that past you to see if you also agree.

model/gurps/skill.go
rinickolous: I'm not sure why I put that there. `excludes` was already not being passed before, but this seems to be erroneous?

ux/skill_editor.go
rinickolous: Yes. Fixed.

ux/skill_editor.go
rinickolous: Yes. Fixed.

model/gurps/skill_default.go
rinickolous: Fixed.

model/gurps/skill.go
rinickolous: Fixed.

model/gurps/skill.go
rinickolous: Fixed.

model/gurps/entity.go
rinickolous: Adjusted to 0.

model/gurps/entity.go
```diff
@@ -1062,7 +1062,7 @@ func (e *Entity) SkillNamed(name, specialization string, requirePoints bool, exc
 // BestSkillMatching returns the highest-level skill whose name and specialization match the given criteria.
 func (e *Entity) BestSkillMatching(nameCriteria, specializationCriteria criteria.Text, replacements map[string]string, requirePoints bool, excludes map[string]bool) *Skill {
 	var best *Skill
-	level := fxp.Min
+	level := fxp.FromInteger(0)
```
richardwilkes: Better to just declare:

```
var level fxp.Int
```

rather than do any computation.

model/gurps/skill.go
richardwilkes: I see. Since there is other stuff that goes along with the SkillDefault data, I wouldn't change it to be different, then. There should probably be code (if there isn't already) that forces the compare value to IsText at load time, to keep it in a consistent state (in case someone edits the file by hand and tries to introduce something else).

richardwilkes (changes requested):

model/gurps/skill.go
rinickolous: Sure. Would that go in `UnmarshalJSON`? I think that's what runs on load time but I'm not 100% sure.

richardwilkes: I'd put the override into `UnmarshalJSONFrom()`

model/gurps/skill.go
```diff
@@ -185,19 +185,18 @@ func skillKind(container bool) byte {
 }
 
 // NewTechnique creates a new technique (i.e. a specialized use of a Skill). All parameters may be nil or empty.
-func NewTechnique(owner DataOwner, parent *Skill, skillName string) *Skill {
+func NewTechnique(owner DataOwner, parent *Skill, skillName criteria.Text) *Skill {
```
richardwilkes: Hmm... since the Compare value is actually fixed, the original API here is probably better, as it doesn't imply you can do things other than exact matches. If you revert this, it means you also don't need to update skill_provider.go to pass in a criteria.Text, which it appears you're not currently doing, which would result in a broken build.

rinickolous: Yeah, I thought so too. Reverted and fixed now.

rinickolous:
Ok, I think everything is actually addressed now. The only thing which may be an issue at this point is the fact that specialization string comparison fields use the term "specialization" rather than "required specialization" but I'll leave you to decide whether it's actually an issue and warrants changing across the board. Let me know.

richardwilkes (approved):

Issue #1044, opened 2026-05-17
Traits don't provide weapons to the scripting context unless they're levelled


maw3193:
While adding scripted notes to some trait modifiers to check they're being used in an apopriate trait (see https://github.com/richardwilkes/gcs_master_library/pull/539), I noticed that the `findWeapons` function was undefined.

Below is a screenshot and a copy of a character sheet that shows the error I experienced, and also two traits with their properties listed in the notes - one has findWeapons and one doesn't.

<img width="815" height="847" alt="Image" src="https://github.com/user-attachments/assets/9fd990ea-09a3-4533-b598-3a8a9d47fb76" />

[Unlevelled trait weapons.gcs.json](https://github.com/user-attachments/files/27902717/Unlevelled.trait.weapons.gcs.json)

I think that the cause is that they're inside the "if trait.CanLevel" condition at https://github.com/richardwilkes/gcs/blob/90b1ce05d138871c421bb66a8e14fe6f2929d30e/model/gurps/scripting_trait.go#L81.

maw3193:
I've created https://github.com/richardwilkes/gcs/pull/1045, which seems to fix the problem.

PR #1045, opened 2026-05-17
Scripting: make Traits provide weapons when they're not levelled

maw3193:
As raised in https://github.com/richardwilkes/gcs/issues/1044.

I'm not a confident Go programmer, but I've tried it out and it looks like it works to me.


richardwilkes (commented):
This seems reasonable. I'm not somewhere I can test this right now, however. I'll try to check it later this week.

richardwilkes (approved):

PR #1050, opened 2026-06-17
Add Fantasy Folk: The Reptilian Races page reference mapping

maw3193:
I'm putting together a pull request for the templates from Fantasy Folk: The Reptilian Races, so I've added a page reference mapping for it.

I omitted a letter for "The" as the convention seems to be to ignore connecting words like "of" and "the", e.g. The Broken Clockwork World is BCW.


richardwilkes:
For changes to the list of known page reference mappings, it is a bit more involved than just this. The master data for this is actually kept in this file:

https://github.com/richardwilkes/gcs_web/blob/master/page_references.txt

Once added there, a simple Go program in that same directory (main.go) is used to produce both the HTML block (for the GCS web site) and the Go source data (for GCS). That way, I know both things are kept in sync from a single master source.

I went ahead and made the changes in the appropriate places and merged them. Note that on the GCS side, I did this in my current development branch, "unison_update", which I'll likely be merging to the main branch soon.

PR #1097, opened 2026-08-17
Refactor with self-referential generics

LeeSaferite:
The original NodeTypes and Node interfaces didn't use self-referential
constraints. I believe this is simply becasue it was written before that
was allowed fully in the 1.26 release. In any case, it's possible to
define a generic type constraint that references itself now. This solves
what seems to be an oddity driven by the previous limitations. With this
update, NodeTypes and Node merge into a single `Node[T Node[T]]`
*constraint*. I highlight "constraint" because with the merge containing
the list of allowed node types, it's limited to being just a constraint.
This does have limitations that need to be respected, but they are all
solvable. One of the larger ones was what to do with the existing type
assertion checks for the node implementations. After a little research I
found that we could have a zero-arg and zero-body generic function that
enforces the type constraint at compile time. Most of the changes in
this PR are mechanical to switch to the new `Node[T Node[T]]`
constraint. There are a few editorial changes made to make the
transition easier. In a few locations (TL, Points, Levels) there were
heavy interfaces for the node that were being augmented with extra
methods. Those were cleaned up to just use a simple interface that was
narrowed to just the methods required for the task. This ends up being
cleaner overall and removes some friction points with the new Node
constraint. With the merge the old `AsNode` method was removed as it's
no longer needed. The UX Node had two properties (`data` and
`dataAsNode`) that collapsed into the same type and value under two
names. I retired the `dataAsNode` in favor of `data` to remove the
redundancy.

A future change might be removing the constraint portions of `Node[T
Node[T]]`, but there are a few wrinkles to making that happen. The
biggest is the `comparable` constraint being needed to properly compare
values. The current constraint where all values are known to be pointers
makes the `comparable` constraint pointless. That being said, if we
removed the allowed nodes constraint, we'd have to solve the comparison
of nodes. I would suggest that the Node interface gets a comparison
method so we can side-step needing a constraint.

### DIFF
--- a/model/gurps/equipment.go
+++ b/model/gurps/equipment.go
@@ -38,11 +38,12 @@ import (
 )
 
 var (
-	_ WeaponOwner            = &Equipment{}
-	_ LeveledOwner           = &Equipment{}
-	_ Node[*Equipment]       = &Equipment{}
-	_ TechLevelProvider      = &Equipment{}
-	_ EditorData[*Equipment] = &EquipmentEditData{}
+	_ = assertNode[*Equipment]
+	_ = assertEditorData[*Equipment]
+
+	_ WeaponOwner       = &Equipment{}
+	_ LeveledOwner      = &Equipment{}
+	_ TechLevelProvider = &Equipment{}
 )
 
 // Columns that can be used with the equipment method .CellData()
--- a/model/gurps/equipment.go
+++ b/model/gurps/equipment.go
@@ -39,7 +39,7 @@ import (
 
 var (
 	_ = assertNode[*Equipment]
-	_ = assertEditorData[*Equipment]
+	_ = assertEditorData[*EquipmentEditData]
 
 	_ WeaponOwner       = &Equipment{}
 	_ LeveledOwner      = &Equipment{}
--- a/model/gurps/equipment.go
+++ b/model/gurps/equipment.go
@@ -38,11 +38,11 @@ import (
 )
 
 var (
-	_ WeaponOwner                   = &Equipment{}
-	_ LeveledOwner                  = &Equipment{}
-	_ Node[*Equipment]              = &Equipment{}
-	_ TechLevelProvider[*Equipment] = &Equipment{}
-	_ EditorData[*Equipment]        = &EquipmentEditData{}
+	_ WeaponOwner            = &Equipment{}
+	_ LeveledOwner           = &Equipment{}
+	_ Node[*Equipment]       = &Equipment{}
+	_ TechLevelProvider      = &Equipment{}
+	_ EditorData[*Equipment] = &EquipmentEditData{}
 )
 
 // Columns that can be used with the equipment method .CellData()
--- a/model/gurps/equipment_modifier.go
+++ b/model/gurps/equipment_modifier.go
@@ -37,7 +37,7 @@ import (
 
 var (
 	_ = assertNode[*EquipmentModifier]
-	_ = assertEditorData[*EquipmentModifier]
+	_ = assertEditorData[*EquipmentModifierEditData]
 
 	_ GeneralModifier = &EquipmentModifier{}
 )
--- a/model/gurps/equipment_modifier.go
+++ b/model/gurps/equipment_modifier.go
@@ -36,9 +36,10 @@ import (
 )
 
 var (
-	_ Node[*EquipmentModifier]       = &EquipmentModifier{}
-	_ GeneralModifier                = &EquipmentModifier{}
-	_ EditorData[*EquipmentModifier] = &EquipmentModifierEditData{}
+	_ = assertNode[*EquipmentModifier]
+	_ = assertEditorData[*EquipmentModifier]
+
+	_ GeneralModifier = &EquipmentModifier{}
 )
 
 // Columns that can be used with the equipment modifier method .CellData()
--- a/model/gurps/node.go
+++ b/model/gurps/node.go
@@ -61,15 +61,15 @@ type Node[T NodeTypes] interface {
 }
 
 // RawPointsAdjuster defines methods for nodes that can have their raw points adjusted must implement.
-type RawPointsAdjuster[T NodeTypes] interface {
-	Node[T]
+type RawPointsAdjuster interface {
+	Container() bool
 	RawPoints() fxp.Int
 	SetRawPoints(points fxp.Int) bool
 }
 
 // SkillAdjustmentProvider defines methods for nodes that can have their skill level adjusted must implement.
-type SkillAdjustmentProvider[T NodeTypes] interface {
-	RawPointsAdjuster[T]
+type SkillAdjustmentProvider interface {
+	RawPointsAdjuster
 	IncrementSkillLevel()
 	DecrementSkillLevel()
 }
--- a/model/gurps/node.go
+++ b/model/gurps/node.go
@@ -31,10 +31,9 @@ type DataOwnerProvider interface {
 	DataOwner() DataOwner
 }
 
-// Node defines the methods required of nodes in our tables.
-//
-// This interface is a "constraint" and cannot be used as a type param, just as a constraint.
-// In practice this just means that it can only be used inside the square brackets of generic methods/types.
+// The type union in the body makes this a constraint, not an ordinary interface: it may only be used as a type
+// parameter's constraint or embedded in another constraint, never as a type. Generic code takes a T Node[T] type
+// parameter and works with T directly.
 type Node[T Node[T]] interface {
 	// These are the limited set of types that can be nodes. New node types *must* be added here
 	*ConditionalModifier | *Equipment | *EquipmentModifier | *Note | *Skill | *Spell | *Trait | *TraitModifier | *Weapon
@@ -62,14 +61,14 @@ type Node[T Node[T]] interface {
 
 func assertNode[T Node[T]]() {}
 
-// RawPointsAdjuster defines methods for nodes that can have their raw points adjusted must implement.
+// RawPointsAdjuster interface for objects that can have their raw points adjusted.
 type RawPointsAdjuster interface {
 	Container() bool
 	RawPoints() fxp.Int
 	SetRawPoints(points fxp.Int) bool
 }
 
-// SkillAdjustmentProvider defines methods for nodes that can have their skill level adjusted must implement.
+// SkillAdjustmentProvider interface for objects that can have their skill level adjusted.
 type SkillAdjustmentProvider interface {
 	RawPointsAdjuster
 	IncrementSkillLevel()
--- a/model/gurps/node.go
+++ b/model/gurps/node.go
@@ -31,6 +31,8 @@ type DataOwnerProvider interface {
 	DataOwner() DataOwner
 }
 
+// Node defines the methods required of nodes in our tables.
+//
 // The type union in the body makes this a constraint, not an ordinary interface: it may only be used as a type
 // parameter's constraint or embedded in another constraint, never as a type. Generic code takes a T Node[T] type
 // parameter and works with T directly.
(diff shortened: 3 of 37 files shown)


model/gurps/equipment.go
```diff
@@ -38,11 +38,12 @@ import (
 )
 
 var (
-	_ WeaponOwner                   = &Equipment{}
-	_ LeveledOwner                  = &Equipment{}
-	_ Node[*Equipment]              = &Equipment{}
-	_ TechLevelProvider[*Equipment] = &Equipment{}
-	_ EditorData[*Equipment]        = &EquipmentEditData{}
+	_ = assertNode[*Equipment]
+	_ = assertEditorData[*Equipment]
```
richardwilkes: This should be EquipmentEditData, not Equipment

model/gurps/equipment_modifier.go
```diff
@@ -36,9 +36,10 @@ import (
 )
 
 var (
-	_ Node[*EquipmentModifier]       = &EquipmentModifier{}
-	_ GeneralModifier                = &EquipmentModifier{}
-	_ EditorData[*EquipmentModifier] = &EquipmentModifierEditData{}
+	_ = assertNode[*EquipmentModifier]
+	_ = assertEditorData[*EquipmentModifier]
```
richardwilkes: This should be EquipmentModifierEditData, not EquipmentModifier

model/gurps/spell.go
```diff
@@ -40,11 +40,12 @@ import (
 )
 
 var (
-	_ Node[*Spell]                    = &Spell{}
-	_ TechLevelProvider[*Spell]       = &Spell{}
-	_ SkillAdjustmentProvider[*Spell] = &Spell{}
-	_ TemplatePickerProvider          = &Spell{}
-	_ EditorData[*Spell]              = &SpellEditData{}
+	_ = assertNode[*Spell]
+	_ = assertEditorData[*Spell]
```
richardwilkes: This should be SpellEditData, not Spell

model/gurps/trait.go
```diff
@@ -45,11 +45,12 @@ import (
 )
 
 var (
+	_ = assertNode[*Trait]
+	_ = assertEditorData[*Trait]
```
richardwilkes: This should be TraitEditData, not Trait

model/gurps/trait_modifier.go
```diff
@@ -36,10 +36,11 @@ import (
 )
 
 var (
-	_ Node[*TraitModifier]       = &TraitModifier{}
-	_ GeneralModifier            = &TraitModifier{}
-	_ LeveledOwner               = &TraitModifier{}
-	_ EditorData[*TraitModifier] = &TraitModifierEditData{}
+	_ = assertNode[*TraitModifier]
+	_ = assertEditorData[*TraitModifier]
```
richardwilkes: This should be TraitModifierEditData, not TraitModifier

ux/container_conversion.go
```diff
@@ -72,9 +66,9 @@ func InstallContainerConversionHandlers[T ConvertableNodeTypes](paneler unison.P
 }
 
 // CanConvertToContainer returns true if the table's current selection has a row that can be converted to a container.
-func CanConvertToContainer[T ConvertableNodeTypes](table *unison.Table[*Node[T]]) bool {
+func CanConvertToContainer[T gurps.Node[T]](table *unison.Table[*Node[T]]) bool {
 	for _, row := range table.SelectedRows(false) {
-		if data := row.Data(); data != nil && data.CanConvertToFromContainer() && !data.Container() {
+		if data, ok := any(row.Data()).(ConvertableContainer); ok && data.CanConvertToFromContainer() && !data.Container() {
```
richardwilkes: This needs an `&& !xreflect.IsNil(data)` after the second `ok`.

ux/container_conversion.go
```diff
@@ -83,33 +77,33 @@ func CanConvertToContainer[T ConvertableNodeTypes](table *unison.Table[*Node[T]]
 
 // CanConvertToNonContainer returns true if the table's current selection has a row that can be converted to a
 // non-container.
-func CanConvertToNonContainer[T ConvertableNodeTypes](table *unison.Table[*Node[T]]) bool {
+func CanConvertToNonContainer[T gurps.Node[T]](table *unison.Table[*Node[T]]) bool {
 	for _, row := range table.SelectedRows(false) {
-		if data := row.Data(); data != nil && data.CanConvertToFromContainer() && data.Container() {
+		if data, ok := any(row.Data()).(ConvertableContainer); ok && data.CanConvertToFromContainer() && data.Container() {
```
richardwilkes: This needs an `&& !xreflect.IsNil(data)` after the second `ok`.

ux/container_conversion.go
```diff
@@ -83,33 +77,33 @@ func CanConvertToContainer[T ConvertableNodeTypes](table *unison.Table[*Node[T]]
...
 // ConvertToContainer converts any selected rows to containers, if possible.
-func ConvertToContainer[T ConvertableNodeTypes](owner Rebuildable, table *unison.Table[*Node[T]]) {
-	before := &containerConversionList[T]{Owner: owner}
-	after := &containerConversionList[T]{Owner: owner}
+func ConvertToContainer[T gurps.Node[T]](owner Rebuildable, table *unison.Table[*Node[T]]) {
+	before := &containerConversionList{Owner: owner}
+	after := &containerConversionList{Owner: owner}
 	for _, row := range table.SelectedRows(false) {
-		if data := row.Data(); data != nil && data.CanConvertToFromContainer() && !data.Container() {
+		if data, ok := any(row.Data()).(ConvertableContainer); ok && data.CanConvertToFromContainer() && !data.Container() {
```
richardwilkes: This needs an `&& !xreflect.IsNil(data)` after the second `ok`.

ux/container_conversion.go
```diff
@@ -119,23 +113,23 @@ func ConvertToContainer[T ConvertableNodeTypes](owner Rebuildable, table *unison
...
 // ConvertToNonContainer converts any selected rows to non-containers, if possible.
-func ConvertToNonContainer[T ConvertableNodeTypes](owner Rebuildable, table *unison.Table[*Node[T]]) {
-	before := &containerConversionList[T]{Owner: owner}
-	after := &containerConversionList[T]{Owner: owner}
+func ConvertToNonContainer[T gurps.Node[T]](owner Rebuildable, table *unison.Table[*Node[T]]) {
+	before := &containerConversionList{Owner: owner}
+	after := &containerConversionList{Owner: owner}
 	for _, row := range table.SelectedRows(false) {
-		if data := row.Data(); data != nil && data.CanConvertToFromContainer() && data.Container() {
+		if data, ok := any(row.Data()).(ConvertableContainer); ok && data.CanConvertToFromContainer() && data.Container() {
```
richardwilkes: This needs an `&& !xreflect.IsNil(data)` after the second `ok`.

model/gurps/node.go
```diff
@@ -60,39 +60,34 @@ type Node[T NodeTypes] interface {
 	CellData(columnID int, data *CellData)
 }
 
+func assertNode[T Node[T]]() {}
+
 // RawPointsAdjuster defines methods for nodes that can have their raw points adjusted must implement.
```
richardwilkes: Replace "nodes" with "objects", since the interface no longer enforces the constraint that it be a Node.

model/gurps/node.go
```diff
@@ -60,39 +60,34 @@ type Node[T NodeTypes] interface {
...
-type RawPointsAdjuster[T NodeTypes] interface {
-	Node[T]
+type RawPointsAdjuster interface {
+	Container() bool
 	RawPoints() fxp.Int
 	SetRawPoints(points fxp.Int) bool
 }
 
 // SkillAdjustmentProvider defines methods for nodes that can have their skill level adjusted must implement.
```
richardwilkes: Replace "nodes" with "objects", since the interface no longer enforces the constraint that it be a Node.

ux/page_list.go
richardwilkes: This, and the one below, should probably be simplified now. Probably worth scanning the code base for similar things elsewhere, too.

model/gurps/node.go
```diff
@@ -31,15 +31,15 @@ type DataOwnerProvider interface {
...
+type Node[T Node[T]] interface {
+	// These are the limited set of types that can be nodes. New node types *must* be added here
 	*ConditionalModifier | *Equipment | *EquipmentModifier | *Note | *Skill | *Spell | *Trait | *TraitModifier | *Weapon
-	nameable.Applier
-	fmt.Stringer
-}
 
-// Node defines the methods required of nodes in our tables.
-type Node[T NodeTypes] interface {
+	// These are the interface mathods each of the above types must implement as a minimum
```
richardwilkes: "mathods" -> "methods"

model/gurps/node.go
```diff
@@ -31,15 +31,15 @@ type DataOwnerProvider interface {
 	DataOwner() DataOwner
 }
 
-// NodeTypes is a constraint that defines the types that may be nodes.
-type NodeTypes interface {
+// Node defines the methods required of nodes in our tables.
+//
+// This interface is a "constraint" and cannot be used as a type param, just as a constraint.
```
richardwilkes: How about:

```
// The type union in the body makes this a constraint, not an ordinary interface: it may only be used as a type
// parameter's constraint or embedded in another constraint, never as a type. Generic code takes a T Node[T] type
// parameter and works with T directly.
```

in place of these two lines?

ux/container_conversion.go
```diff
@@ -10,49 +10,43 @@
...
-	"github.com/richardwilkes/gcs/v5/model/nameable"
 	"github.com/richardwilkes/unison"
 )
 
-// ConvertableNodeTypes defines the types that the container conversion can work on.
-type ConvertableNodeTypes interface {
-	*gurps.Equipment | *gurps.Note
-	fmt.Stringer
-	nameable.Applier
+// ConvertableContainer defines the types that the container conversion can work on.
```
richardwilkes: Let's change this to read:

```
// ConvertableContainer defines the methods a node must implement to be convertible to and from a container. Rows are
// matched against it at runtime, so nodes that do not implement it are simply skipped.
```

richardwilkes (changes requested):
Overall, looks pretty good. I noted a few things I'd like changed.

ux/container_conversion.go
LeeSaferite: Stupid typed-nils.

model/gurps/node.go
LeeSaferite: I'm not sure I follow your meaning? They are still `Node` types.

LeeSaferite: Ah, the highlight was on the wrong lines. I got it.

ux/page_list.go
LeeSaferite: I don't know the UI well enough, would removing the filter here cause the UI to change for the cases where it's not a cointainer-capable node?

richardwilkes: These handlers are only applicable to Equipment and Notes, so need to be restricted to just those types of lists.

model/gurps/node.go
```diff
@@ -31,10 +31,9 @@ type DataOwnerProvider interface {
 	DataOwner() DataOwner
 }
 
-// Node defines the methods required of nodes in our tables.
```
richardwilkes: Didn't mean for you to delete this line.

LeeSaferite:
@richardwilkes I resolved all but one of the feedback items. The remaining one I'm not sure what change should be made and I don't want to potentially mess up the UI by making a wrong change there. If you have guidance on exactly what you'd like to see changed, I can get it done.

model/gurps/node.go
LeeSaferite: Ah, you said "two". My mistake.

ux/page_list.go
richardwilkes: It looks like this could safely be boiled down to just:
```
InstallContainerConversionHandlers(p, owner, p.Table)
```
since the only things that call that function are the equipment and notes lists. Might be better to eliminate this function entirely and inline that at the call sites.

LeeSaferite: I cleaned it up. Now `InstallContainerConversionHandlers` is smart and only runs when the underlying node type implements the `ConvertableContainer` interface.

richardwilkes (approved):

PR #1098, opened 2026-08-17
Auto-redirect `B` page refs for "Campaigns" to `BX`

LeeSaferite:
For users that have split Basic Set PDFs, any `B` page ref that's pointing to pages in the second PDF (Campaigns) has no way to click those links without editing the library data. After researching and discovering apparently 16k+ references that trigger this (many clickable), I decided to add a remap from B338+ over to BX instead. This happens before the code to PDF lookup, so it properly triggers the page mapping UI for the `BX` code the first time.

This is currently the only case where a single "book" is split across multiple physical books but has a single page number space.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ## New & Improved
 
-- (none yet)
+- Page references for `B:338+` (Basic Set: Campaigns) are redirected to use `BX` automatically. This allows split PDF users to open `B338+` page references. (#1098)
 
 ## Bug Fixes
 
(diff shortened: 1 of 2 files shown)


richardwilkes:
Please add a short release note entry to the "New & Improved" section of the RELEASE_NOTES.md file detailing this change, ideally from the user's point of view.

richardwilkes:
I should have mentioned that the release note entry should have a trailing ` (#1098)` so that Github generates a link to this PR for those that are curious enough to click on it.

richardwilkes (approved):

PR #1100, opened 2026-08-18
Trait/Equipment Modifiers lose Source on Clone

LeeSaferite:
This fixes attached Trait/Equiment modifiers losing their source when the Trait or Equipment is cloned.

The previous behavior was selecting the wrong library path. By using `one` it was referencing the *origin* path, when clone is expecting the *destination* path. The passed in trait/equipment has already had it's source fixed by the time Clone is called on each modifier. This lets us use the source library path from the passed in trait/equipment to get the correct value.

The net effect of this change is that modifiers attached to traits and equipment get a proper source value when they are copied over to character sheets and templates.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -46,3 +46,4 @@
   focus was returned to the list in a way that scrolled the entire list into view, which on a sheet with a long list
   could move the row far from where it had been. Now only the edited row is brought into view, and only when it isn't
   already visible.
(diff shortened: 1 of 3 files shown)


LeeSaferite:
If the modifier being cloned already has a value source, the existing code just leaves it alone. This change only kicks in when the modifier being copied has no source data. At that point the modifier being cloned is considered the source.

richardwilkes (approved):

PR #1101, opened 2026-08-19
Trigger modifier selections and substitutions for templates

LeeSaferite:
This changes the behavior of to "Copy To Template" and the drag to template operations. Previously they did not trigger the modifier config or the substitions config.

### DIFF
--- a/ux/table.go
+++ b/ux/table.go
@@ -356,12 +356,12 @@ func copySelectionToSheet[T gurps.Node[T]](table *unison.Table[*Node[T]]) {
 // then folding the points of rows that duplicate ones already present into those rows. Does nothing when the
 // destination isn't a character or loot sheet.
 func processCopiedRowsForSheet[T gurps.Node[T]](source, target *unison.Table[*Node[T]]) {
-	if !isForCharacterOrLootSheet(target) {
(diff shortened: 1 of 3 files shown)


ux/table.go
```diff
@@ -356,12 +356,12 @@ func copySelectionToSheet[T gurps.Node[T]](table *unison.Table[*Node[T]]) {
 // then folding the points of rows that duplicate ones already present into those rows. Does nothing when the
 // destination isn't a character or loot sheet.
 func processCopiedRowsForSheet[T gurps.Node[T]](source, target *unison.Table[*Node[T]]) {
-	if !isForCharacterOrLootSheet(target) {
+	if !isForCharacterOrLootSheet(target) && !isForTemplate(target) {
```
richardwilkes: Since you're adding the isForTemplate part to every use of isForCharacterOrLootSheet, let's just incorporate the two together (and probably give it a better name, maybe `shouldProcessModifiersAndNameables`?).

LeeSaferite: I originally had that, but in an effort to be less traumatic on changes I went this route. I 100% do not mind changing though.

richardwilkes (approved):

PR #1102, opened 2026-08-19
Add 'preconfigured' flag for template items

LeeSaferite:
Current template behavior is that when applying a template to a character, all modifier and substition config is done for each template item as it's applied. In most cases the template creator has already made modifier and substitution selections when creating the template. This causes a poor UX when applying a template. It opens the user up to confusion about "should I select modifiers" and "what's this subtitution meant to cover?"

This commit adds a new, optional, flag to template traits, skills, spells, and notes. When this flag is off or missing a template acts exactly the same as it has until now. When a template designer is creating a template they can choose to flag some or all items as "preconfigured" to signal that the modifier and substitution processing during template application can be skipped. If a template item has the flag set but some substitutions are *not* set, then the ones that are not set will be prompted during processing while the already-set ones are skipped.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,7 +12,6 @@
   don't apply.
 - Spells can now have features, just like traits, skills and equipment. As a side effect, weapons attached to a spell
   now receive any "this weapon" bonuses defined by the spell's own features.
-- Template items (non-container) now have a "Preconfigured" flag used to indicate that modifier and substitution configuration can be skipped when adding thie item to a character. (#1102)
 - Adding items to a template now triggers modifier selections and nameable replacements (#1101)
 - Template items (non-container) now have a "Preconfigured" flag used to indicate that modifier and substitution configuration can be skipped when adding thie item to a character. (#1102)
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,6 +14,7 @@
   now receive any "this weapon" bonuses defined by the spell's own features.
 - Template items (non-container) now have a "Preconfigured" flag used to indicate that modifier and substitution configuration can be skipped when adding thie item to a character. (#1102)
 - Adding items to a template now triggers modifier selections and nameable replacements (#1101)
+- Template items (non-container) now have a "Preconfigured" flag used to indicate that modifier and substitution configuration can be skipped when adding thie item to a character. (#1102)
 
 ## Bug Fixes
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,7 @@
   don't apply.
 - Spells can now have features, just like traits, skills and equipment. As a side effect, weapons attached to a spell
   now receive any "this weapon" bonuses defined by the spell's own features.
+- Template items (non-container) now have a "Preconfigured" flag used to indicate that modifier and substitution configuration can be skipped when adding thie item to a character. (#1102)
 
 ## Bug Fixes
 
--- a/model/gurps/equipment.go
+++ b/model/gurps/equipment.go
@@ -1204,3 +1204,8 @@ func (e *EquipmentEditData) copyFrom(equipment *Equipment, other *EquipmentEditD
 	e.Weapons = CloneWeapons(other.Weapons, equipment, isApply)
 	e.Features = other.Features.Clone()
 }
+
+// CanPreconfigureContainer implements Preconfigurable.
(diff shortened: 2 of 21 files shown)


model/gurps/equipment.go
```diff
@@ -1203,3 +1204,8 @@ func (e *EquipmentEditData) copyFrom(equipment *Equipment, other *EquipmentEditD
 	e.Weapons = CloneWeapons(other.Weapons, equipment, isApply)
 	e.Features = other.Features.Clone()
 }
+
+// CanPreconfigureContainer implements Preconfigurable.
+func (e *EquipmentEditData) CanPreconfigureContainer() bool {
```
richardwilkes, line 1209: I'm a little concerned about both this override and the same thing you have for traits, as I've found it can be rather easy to call the underlying one and therefore get the wrong response. It helps that you made preconfigurable non-public, but only for those trying to access it out of this package.

I don't see an easy way to achieve the same effect without having data storage, which then has to be initialized, so lets leave it for now. Just wanted to raise the issue so you were aware that I'd like to avoid such things when possible.

richardwilkes: Hmm... although we could make this work without the override and without data storage if we passed in the owning object and switched on its type. I think I'd prefer that, even though it is slightly more expensive, unless you find that there is some reason we can't pass in the object it is embedded in to achieve this.

ux/table_drop.go
```diff
@@ -192,6 +192,9 @@ func didDropCallback[T gurps.Node[T]](undo *unison.UndoEdit[*TableDragUndoEditDa
 			MergeAddedRows(to)
 		}
 	}
+	if clearPreconfiguredFlag(to, nil) {
+		to = liveTable(to)
```
richardwilkes, line 196: clearPreconfiguredFlag() doesn't rebuild the table, so why is it necessary to do a `to = liveTable(to)` when it returns true?

RELEASE_NOTES.md
```diff
@@ -12,7 +12,9 @@
   don't apply.
 - Spells can now have features, just like traits, skills and equipment. As a side effect, weapons attached to a spell
   now receive any "this weapon" bonuses defined by the spell's own features.
+- Template items (non-container) now have a "Preconfigured" flag used to indicate that modifier and substitution configuration can be skipped when adding thie item to a character. (#1102)
 - Adding items to a template now triggers modifier selections and nameable replacements (#1101)
+- Template items (non-container) now have a "Preconfigured" flag used to indicate that modifier and substitution configuration can be skipped when adding thie item to a character. (#1102)
```
richardwilkes, line 16: Lets remove this redundant entry and modify the earlier one to either remove the parenthetical or be a bit more specific if we aren't actually showing the checkbox for some item editors.

ux/table_drop.go
LeeSaferite: Over-abundance of caution. It's changing the data on the underlying rows of the `to` table so I wasn't sure if there was an impact that needed to be accounted for. Hence to decision to call `tp = liveTable(to)`. If you say it's not needed, I'm fin removing the call.

RELEASE_NOTES.md
LeeSaferite: Gah. Not sure how it got doubled. Sorry about that.

model/gurps/equipment.go
LeeSaferite: In this case, I'm not even sure *how* you could call the underlying default method. We are operating of of an EditorData interface and type checking that to the `Preconfigured` interface. But it's all still the underlying `EquipmentEditData` struct. When the call to `CanPreconfigureContainer` happens against that concrete struct, you get the override. The only way to get the default would be to somehow directly access the `preconfigured` embed and them call `CanPreconfigureContainer` directly on that. That's highly unlikely. It's practically impossible in the code that's actually using it as it's another package. You'd have to try *really* hard to make it happen. But I will concede that I could be mistaken.

I just wanted to invert it like this so the caller didn't need to check for specific node types and could just depend on the interface methods to discover what it needed instead. I think there are a lot of places int he code that could benefit from similar treatment so we could not embed specific node type checks and replace them with capability checks instead.

In any case, if you prefer we approach this a different way, I'm game. Just let me know how you want to handle this.

richardwilkes: We're just going to keep this as-is.

ux/table_drop.go
richardwilkes: Pretty sure this isn't needed... but it is also effectively a noop, so fine to leave

richardwilkes (approved):

PR #1111, opened 2026-08-24
Template Choices config not saving

LeeSaferite:
When editing a container, changing the template choices config wouldn't mark the record as change and the changes would not save as expected.

The issue was a simple mistaken use of `e.target` vs `e.editorData`.

During this exploration I also discovered that the spell container sync data was using the one from skills. While they are identical, this was prone to future issues.

I also switched to an embedded templatePicker implementation that's now shared by Traits, Skills, and Spells.


richardwilkes:
What's the point of the indirection through templatePicker?

richardwilkes (approved):
Thanks

PR #1113, opened 2026-08-27
Adding new capabilities for nameable markers

LeeSaferite:
The existing nameable format was simply `@<Label>@`. While this is functional, it was obviously an issue when you audited the master library data for nameable markers. The formats are all over the place and needed some structured additional capability to help users.

This new feature adds several capabilities that are backwards compatible and essentially just UI-side improvements. The new format continues to support a `Label`, but it adds the ability to pre-define `Option` values, flag the marker as accepting free-form text entry, flag the marker as accepting empty text, and adding a tooltip. Additionally, now every render of the unresolved nameable uses just the label. The only time the full nameable marker text is shown is when the marker text is in a field to edit. This is called the "compact" view and makes markers take up less space. This feature adds a new editable combobox widget that's used to make all of this happen. Using this widget we also support a special-case combo selection to indicate "not set". This removed the marker from the replacements map on the item. The effect is that if the item is ever copied or applied to a template or character, the substitutions popup will trigger again for that marker. Lastly, we apply a very conservative regex detect against old marker text to attempt to extract more structured data from the marker for display.

Examples of the new format:
- `@Label@` - Same as old behavior, free-form text, empty allowed
- `@Label|*@` - Just free-form text, empty not allowed
- `@Label|*|?@` - Free form text and empty are both valid
- `@Label|One|Two@` - Mandatory selection of "One" or "Two"
- `@Label|One|Two|?@` - Mandatory selection of "One" or "Two" or the `«empty»` option
- `@Label|One|Two|?|*@` - Selection of "One" or "Two" or the `«empty»` option, or free-form text
- `@Label|tt(This is a tooltip)|One|Two|?@` - Includes a tooltip


model/nameable/marker.go
```diff
@@ -0,0 +1,125 @@
...
+	AllowEmpty bool
+	FreeForm   bool
+}
+
+// splitSegments splits raw on '|', honoring '\|' as an escaped, literal pipe rather than a delimiter, and '\n' as
+// an escaped newline.
+func splitSegments(raw string) []string {
+	segments := make([]string, 0, strings.Count(raw, "|")+1)
+	var buf strings.Builder
+	for i := 0; i < len(raw); i++ {
```
richardwilkes: This loop is safe for what we're doing here, but if we add other values we're checking for in the future, we need to be aware that this is processing byte-by-byte and not rune-by-rune.

model/nameable/marker.go
```diff
@@ -0,0 +1,125 @@
...
+// defined by the Mozilla Public License, version 2.0.
+
+package nameable
+
+import (
+	"errors"
+	"strings"
+)
+
+// Errors returned by ParseSyntax explaining why raw could not be parsed.
```
richardwilkes: Please use the github.com/richardwilkes/toolbox/v2/errs package instead of sentinel errors, so that stack traces are emitted with the errors.

richardwilkes (changes requested):
I have some feedback on the combo field I'll give interactively in chat.

model/nameable/nameable.go
```diff
@@ -41,19 +41,41 @@ func Extract(str string, m, existing map[string]string) {
...
+			parts := strings.Split(str, "@")
+			var b strings.Builder
+			for i, one := range parts {
+				if i%2 == 1 && i < count {
+					b.WriteByte('@')
+					b.WriteString(ParseMarker(one).Label)
+					b.WriteByte('@')
+					continue
+				}
+				b.WriteString(one)
```
richardwilkes: In the case where the user had something broken in here, like `Contact @Who@ at admin@site.com`, this will turn the last bit into `adminsite.com`, dropping the  @, since it isn't paired properly.

model/nameable/marker.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return marker, nil
+}
+
+// DefaultValue returns the value that should be used for key when no explicit replacement has been recorded for it
+// yet. For a key using the pipe-delimited syntax, this is the empty string when AllowEmpty is set (empty is itself
+// a valid value there) or its first option otherwise. For any other key, this is the key itself, matching the
+// long-standing behavior of showing the raw key text until the user provides a value.
+func DefaultValue(key string) string {
+	if marker, err := ParseSyntax(key); err == nil {
```
richardwilkes: For `@Label|*@` (freeform, no options, no `?`) `DefaultValue` returns `""`, which it is defined not to... although I'm not sure what else you can reasonably return.

model/nameable/marker.go
```diff
@@ -0,0 +1,125 @@
...
+	for _, part := range parts[1:] {
+		switch {
+		case part == AllowEmptyToken:
+			marker.AllowEmpty = true
+		case part == FreeFormToken:
+			marker.FreeForm = true
+		case strings.HasPrefix(part, tooltipPrefix) && strings.HasSuffix(part, tooltipSuffix):
+			tooltipLines = append(tooltipLines, strings.Split(part[len(tooltipPrefix):len(part)-len(tooltipSuffix)], "\n")...)
+		default:
+			marker.Options = append(marker.Options, part)
```
richardwilkes: Should probably use `strings.TrimSpace()` to eliminate unwanted leading & trailing spaces in the options, which you'll get if you do something like:

`@label | foo | bar @`

Probably should do that for the label, too.

ux/nameables_processing.go
```diff
@@ -139,14 +139,18 @@ func ShowNameablesDialog(titles []string, nameables []map[string]string, visible
 		})
 		list.AddChild(header)
 		for _, k := range keys {
+			marker := nameable.ParseMarker(k)
 			label := unison.NewLabel()
-			label.SetTitle(k)
+			label.SetTitle(marker.Label)
```
richardwilkes: The list is de-duped by the full key, but the display then only shows the label portion... which can result in more than one entry that looks identical to the user unless they carefully compare the popup options available in each. Not sure there is much we can do about that, but wanted to raise it in case you had a brilliant idea. The sort order in that case, too, may seem random to the user, as sorting is done above on the full key, not on the actual title used.

LeeSaferite: I was thinking about this and the thing I can up with is adding mor context to where the nameable originated, not just the top-level trait for instance, but if it's a modifier, add breadcrumbs. The *could* help disambiguate, maybe.

model/nameable/extract.go
```diff
@@ -0,0 +1,249 @@
...
+				start = i + cl
+			}
+
+			// Advance to the next rune
+			continue
+		}
+
+		// A begin delimiter inside a placeholder is a special case.
+		// It's treated like a misidentified placeholder start.
+		if r == begin && isPlaceholder {
```
richardwilkes, line 120: This can't actually happen, given that the start & end markers are the same rune.

model/nameable/marker.go
```diff
@@ -0,0 +1,224 @@
...
+	FreeFormToken      string = "*"
+	TooltipPrefixToken string = "tt("
+	TooltipSuffixToken string = ")"
+)
+
+var (
+	// legacyExampleListPattern matches an old-format marker shaped like "Label: item, item[, item...], etc." -- a
+	// short leading label, then two or more comma-separated example items, ending in a trailing "etc."/"etc" that
+	// signals the list isn't exhaustive. The item list itself is captured without the trailing ", etc." part.
+	legacyExampleListPattern = regexp.MustCompile(`^([^:\n]{1,40}): ((?:[^,\n]+, )+[^,\n]+)(, ?etc\.?)$`)
```
richardwilkes, line 35: While we may truncate the display of the label to 40 characters, we shouldn't ignore a legacy substitution that used more. This applies to legacyLabeledPattern below, too.

model/nameable/marker.go
```diff
@@ -0,0 +1,224 @@
...
+	if label == "" {
+		return Marker{Raw: raw}, false
+	}
+
+	if len(segments) == 1 {
+		if m := legacyExampleListPattern.FindStringSubmatch(label); m != nil {
+			var options []string
+			for item := range strings.SplitSeq(m[2], ", ") {
+				if item = strings.TrimSpace(item); item != "" {
+					options = append(options, item)
```
richardwilkes, line 64: Should we ensure the option is unique before placing it in the list here?

model/nameable/marker.go
```diff
@@ -0,0 +1,224 @@
...
+			continue
+		}
+		if s == AllowEmptyToken {
+			allowEmpty = true
+			continue
+		}
+		if strings.HasPrefix(s, TooltipPrefixToken) && strings.HasSuffix(s, TooltipSuffixToken) {
+			tooltipSegment := s[len(TooltipPrefixToken) : len(s)-len(TooltipSuffixToken)]
+			tooltipSegment = strings.TrimSpace(UnescapeRunes(tooltipSegment, EscapeRune, '\n'))
+			tooltips = append(tooltips, strings.Split(tooltipSegment, "\n")...)
```
richardwilkes: In the docs, did you note that the tooltip wrapper could be on multiple segments, that they would be merged together into a single tooltip with a return in between, and that they didn't have to be in the second position (like the examples that I saw show)? All good capabilities, just was surprised to see this.

model/nameable/marker.go
```diff
@@ -0,0 +1,224 @@
...
+	if m.Tooltip != "" {
+		sb.WriteRune(SegmentDelimiter)
+		sb.WriteString(TooltipPrefixToken)
+		sb.WriteString(EscapeRunes(m.Tooltip, EscapeRune, SegmentDelimiter, '\n'))
+		sb.WriteString(TooltipSuffixToken)
+	}
+	return sb.String()
+}
+
+// DefaultValue return the default value for this marker when no replacement is provided
```
richardwilkes: return -> returns

ux/nameables_processing.go
```diff
@@ -139,14 +145,27 @@ func ShowNameablesDialog(titles []string, nameables []map[string]string, visible
...
 		for _, k := range keys {
+			marker, ok := nameable.NewMarker(k)
+			if !ok {
+				continue
+			}
 			label := unison.NewLabel()
-			label.SetTitle(k)
+			title := xstrings.Truncate(marker.Label, 60, true)
+			tooltip := marker.Tooltip
+			if strings.HasSuffix(title, "…") {
```
richardwilkes: You should do `marker.Label != title` instead, since a label might legitimately end with an ellipsis.

ux/nameables_processing.go
```diff
@@ -174,17 +193,72 @@ func ShowNameablesDialog(titles []string, nameables []map[string]string, visible
...
+		m[marker.Key()] = *value
+	}
+
+	if marker.FreeForm {
+		field := unison.NewComboField(options, initial, apply)
+
+		// Use a fixed string to set the lower bounds on the ComboField width
+		minWidth := field.MinimumTextWidth
+		field.SetMinimumTextWidthUsing("A reasonably wide string")
+		field.MinimumTextWidth = max(field.MinimumTextWidth, minWidth)
```
richardwilkes, line 228: This is unnecessary -- SetMinimumTextWidthUsing is already doing the work

richardwilkes (changes requested):

model/nameable/marker.go
LeeSaferite: We aren't ignoring the substitution. This is just a safety limit on the legacy regex match. If it exceeds the limit it doesn't attempt to parse the legacy marker, but it still works fine. The "label" will be the full marker text. Basically it's the old behavior.

model/nameable/marker.go
LeeSaferite: Updated.

ux/nameables_processing.go
LeeSaferite: Resolved.

ux/nameables_processing.go
LeeSaferite: This is ensuring a minimum width for the widget, regardless of the width of the pre-defined options. It's making sure there's enough space for typing. Without setting the min width like this, we just get the width of our string. If an option is longer we don't expand. If you have a better method for this, please advise.

model/nameable/extract.go
LeeSaferite: Yes, but I wrote the ExtrartParts code to work with matching delimiters or different begin/end markers. For a single delimiter case you are correct, that state cannot happen.

ux/nameables_processing.go
richardwilkes: My point was that you only need the field.SetMinimumTextWidthUsing() call. The other two parts are pointless.

model/nameable/marker.go
LeeSaferite: I do not recall TBH, but I can revisit and make sure it's there.

ux/nameables_processing.go
LeeSaferite: We just discussed this on discord, but I'll recap here. The call to `NewComboField` is a widget builder and it's already setting a minimum width based on the items in the options list. This section of code is saying we want the field to have a minimum width that is the larger of the initial min based on option or a static exemplar string. This is to ensure the field isn't unreasonably narrow so someone has room to type out their value without in-field scrolling.

model/nameable/marker.go
LeeSaferite: The user guide text does call out they can be anywhere and implies multiplicity but doesn't explicitly state it is supported. I'm inclined to leave the guide as-is unless you feel strongly otherwise.

richardwilkes (approved):

PR #1114, opened 2026-08-27
Feature: Combine Ruleset

JackChance:
Creates a new option on the File... drop down for combining rulesets into a single, user-named ruleset
Saves custom ruleset to the User Library location

Allows for GMs to create a unified ruleset to distribute using the prioritized list of books for their setting. Solves the issue of having many potential sources of truth for a given campaign (Is this equipment in the setting specific book or do we have to default to Basic Set for this one?) and maintains the page references for all instances of objects collapsed during the combination process.


richardwilkes:
Thanks for the work on this. I'm not yet convinced GCS needs this feature, but setting that aside, there are several things that would need attention before it could be considered. First, the branch no longer compiles after the merge of main: `combine_rulesets.go` still calls `gurps.GlobalSettings().Libraries()` as a method, which the main branch turned into a field (the fix is just dropping the parentheses in two places). Second, and the biggest design question: because rows are cloned in Reference mode, items dragged from a combined file onto a sheet end up syncing against the *original book files*, not the combined ruleset. "Sync with Source" will then restore the book's stats and discard the merged page references, which seems like the opposite of the feature's intent. Related, every regeneration mints new IDs for all rows, so nothing can hold a stable link to a combined file across re-runs (the rules lookup download already solves this by reusing IDs from the existing file). It's worth deciding explicitly whether the combined file is meant to be the campaign's source of truth and picking the clone/ID strategy to match.

There are also a number of smaller problems. The merge behavior has some surprises: the "Include subfolders" checkbox changes which duplicate wins, duplicates within a single source file get merged away, and a mid-run failure leaves a half-written folder behind. The UI needs polish too — the combine runs synchronously on the UI thread, the dialog opens far too tall, and the name field silently rewrites some characters while letting Windows-invalid names through. Finally, the tests are thin: only 3 of the 7 data types are covered and the subfolder path is never exercised.

One more note: as with the Master Library, using generative AI to prepare a contribution isn't forbidden here, but the same expectations apply. Please say which parts were AI-generated, and take as much responsibility for the submission as if you'd written it all yourself — that means having read and understood every change, and having built and exercised the result before asking for review. AI can produce a large volume of plausible-looking code whose errors are time-consuming for a reviewer to dig out.

JackChance:
Thanks for the feedback. I'll cycle on this a bit longer, solve the issues you raised and try to head off any further ones.

PR #1115, opened 2026-08-27
Switch from omitzero to omitempty for maps and slices

LeeSaferite:
Using `omitzero` will still output an empty slice or map. Switching to `omitempty` cause a slcie/map to omit rendering in JSON when it's nil or empty.


richardwilkes (approved):
Once CI finishes successfully, feel free to merge it.

PR #1118, opened 2026-09-01
Adding Equipment Template Choices

LeeSaferite:
Traits, Skills, and Spells already support tempalte choices. This is adding the ability to set an equipment container to a template choice container. This means a user selects base don points or items and then the selected items are added to teh character sheet while the group is not.


richardwilkes:
This is incomplete. Equipment.CellData never shows the "Pick N" tag and the template picker only supports counts or points -- equipment has no points and might instead need a quantity, cost, or maybe even weight criteria instead.

LeeSaferite:
Closing the PR until I come back with one that covers equipment better.

PR #1121, opened 2026-09-01
Refactor template picker usage

LeeSaferite:
The existing 'TemplatePicker' struct and 'TemplatePickerProvider' interface have several issues, the biggest being that there's no way to define a set of allowed picker types based on the containing node. This refactor changes from using direct access to the 'TemplatePicker' struct to having the interface capable of returning an allowed list of types and a reference to the underlying 'TemplatePicker' data. The interface also contractually promises that the returned 'TemplatePicker' will never be nil due to the backing `TemplatePicker` no longer being a pointer. This allowed for the remove of a lot of code that was there just to deal with `TemplatePicker` being nilable. The json output is there same due to the use of the `omitzero` flag and TemplatePicker.IsZero()`.

### DIFF
--- a/model/gurps/template_picker.go
+++ b/model/gurps/template_picker.go
@@ -20,10 +20,10 @@ import (
 	"github.com/richardwilkes/toolbox/v2/xhash"
 )
 
-// TemplatePickerProvider defines the methods a TemplatePicker provider has.
+// TemplatePickerProvider provides the method used to get a list of valid picker type and to access the picker data
 type TemplatePickerProvider interface {
-	TemplatePickerData() *TemplatePicker
-	SetTemplatePickerData(*TemplatePicker)
+	// TemplatePickerData returns a list of valid picker types and valid, non-nil, pointer to a TemplatePicker struct
+	TemplatePickerData() ([]picker.Type, *TemplatePicker)
 }
 
 // TemplatePicker holds the data necessary to allow a template choice to be made.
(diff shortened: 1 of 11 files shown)


model/gurps/template_picker.go
```diff
@@ -65,7 +56,7 @@ func (t *TemplatePicker) String() string {
 }
 
 // Hash writes this object's contents into the hasher.
-func (t *TemplatePicker) Hash(h hash.Hash) {
+func (t TemplatePicker) Hash(h hash.Hash) {
 	xhash.Num8(h, t.Type)
 	t.Qualifier.Hash(h)
```
richardwilkes: This should not emit anything for fields other than the Type if the Type is NotApplicable. Otherwise, the hash ends up with garbage values being incorporated, which makes them not match if you've done any editing.

ux/choices.go
```diff
@@ -32,20 +32,17 @@ func addChoices[N gurps.Node[N], D gurps.EditorData[N]](e *editor[N, D], parent
...
 	}
 
-	if tp == nil {
-		tp = &gurps.TemplatePicker{}
-	}
-
 	last := tp.Type
 	wrapper := addFlowWrapper(parent, i18n.Text("Choices"), 3)
-	typePopup = addPopup(wrapper, picker.Types, &tp.Type)
+	typePopup = addPopup(wrapper, types, &tp.Type)
```
richardwilkes, line 45: Since the Type might be a choice that isn't in the returned types list (it is loaded from a file, which could have old information or just info that was directly edited and possibly invalid), should force Type to be one of the valid choices before passing it to addPopup -- otherwise the initial selection won't be performed.

richardwilkes: Would probably be best to do this enforcement at load time rather than here.

LeeSaferite: I would say both are correct. What behavior should we take if it's an unknown type? Just set it back to the default value?

LeeSaferite: Just reviewed the enum code, it's already impossible to get an invalid picker type from the json data. If the type string doesn't map to a known value, it reverts to the default value. So, barring us writing bad code, it is impossible to have an unknown picker type.

richardwilkes: The issue isn't that the type is invalid -- it's that it may not be one of the actually permitted choices for that item. Can't happen if the user only uses GCS to edit things, but many folks do direct editing, especially for bulk changes.

richardwilkes (approved):

PR #1122, opened 2026-09-01
Adding Equipment Template Choices

LeeSaferite:
Traits, Skills, and Spells already support template choices. This is adding the ability to set an equipment container to a template choice container. This means a user selects based on count, quantity, cost, or weight and then the selected items are added to the character sheet while the template choice group is not added.

This builds on PR (#1121)


LeeSaferite:
The UI strings/descriptions of the template picker have a commit pending a push that are waiting on  #1123 to be merged. This will allow the descriptions to be unit-aware.

PR #1125, opened 2026-09-03
Adding Alternative Ability Slots

LeeSaferite:
The trait alternative ability container only supports a single alternative ability "slot" currently. The rules mention that you can have multiple slots. This adds a slot configuration, adjusts the table to indicate the slot count, and adjusts the point calculations for the container.

### DIFF
--- a/model/gurps/trait.go
+++ b/model/gurps/trait.go
@@ -10,6 +10,7 @@
 package gurps
 
 import (
+	"cmp"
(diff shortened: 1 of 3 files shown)


richardwilkes:
For existing traits that use this trait container type, they get a field marked invalid with a 0 in it when opened. These should probably be transformed automatically to a 1.

richardwilkes (approved):

PR #1127, opened 2026-09-05
Refactoring Modifier Nodes

LeeSaferite:
The current code around modifiers for traits and equipment is very similar and we can start unifying this code by good use of generics.

I have added a new `modifiers.go` file in the `gurps` package to hold interfaces, constraints, and any common code. That involvde migrating the existing `GeneralModifer` interface from `trait.go` where it was initially created. I added two new constraint and matching "assert" helpers to verify compliance to the constraints at compile time. There's also a shared 'modifiable' implementation that uses generics to work with both equipment and traits.

Future work will cover further identification and refactoring of code that's identical between equipment and trait modifiers.

### DIFF
--- a/model/gurps/equipment.go
+++ b/model/gurps/equipment.go
@@ -42,6 +42,7 @@ import (
 var (
 	_ = assertNode[*Equipment]
 	_ = assertEditorData[*EquipmentEditData]
+	_ = assertModifiable[*EquipmentEditData]
 
 	_ WeaponOwner       = &Equipment{}
 	_ LeveledOwner      = &Equipment{}
@@ -89,14 +90,18 @@ type EquipmentData struct {
 // EquipmentEditData holds the Equipment data that can be edited by the UI detail editor.
 type EquipmentEditData struct {
 	EquipmentSyncData
-	VTTNotes     string               `json:"vtt_notes,omitzero"`
-	Replacements map[string]string    `json:"replacements,omitempty"`
-	Modifiers    []*EquipmentModifier `json:"modifiers,omitempty"`
-	RatedST      fxp.Int              `json:"rated_strength,omitzero"`
-	Quantity     fxp.Int              `json:"quantity"`
-	Level        fxp.Int              `json:"level,omitzero"`
-	Uses         int                  `json:"uses,omitzero"`
-	Equipped     bool                 `json:"equipped,omitzero"`
+	VTTNotes     string            `json:"vtt_notes,omitzero"`
+	Replacements map[string]string `json:"replacements,omitempty"`
+
+	// Leverage a default generic implementation of the Modifiable constraint that makes this a ModifiableNode
+	modifiable[*EquipmentModifier, *Equipment]
+
+	RatedST  fxp.Int `json:"rated_strength,omitzero"`
+	Quantity fxp.Int `json:"quantity"`
+	Level    fxp.Int `json:"level,omitzero"`
+	Uses     int     `json:"uses,omitzero"`
+	Equipped bool    `json:"equipped,omitzero"`
+
 	ItemSwitch
 	preconfigurable
 }
@@ -623,7 +628,7 @@ func (e *Equipment) SetDataOwner(owner DataOwner) {
 		}
 	}
 	for _, m := range e.Modifiers {
-		m.setEquipment(e)
+		m.SetTargetNode(e)
 		m.SetDataOwner(owner)
 	}
 }
@@ -883,7 +888,7 @@ func ContainedWeightAdjustedForModifiers(equipment *Equipment, defUnits fxp.Weig
 		}
 	}
 	Traverse(func(mod *EquipmentModifier) bool {
-		mod.setEquipment(equipment)
+		mod.SetTargetNode(equipment)
 		for _, f := range mod.Features.Active(switchedOn) {
 			if cwr, ok := f.(*ContainedWeightReduction); ok {
 				if cwr.IsPercentageReduction() {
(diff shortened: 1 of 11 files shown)


model/gurps/equipment.go
```diff
@@ -89,14 +90,18 @@ type EquipmentData struct {
...
-	Modifiers    []*EquipmentModifier `json:"modifiers,omitempty"`
-	RatedST      fxp.Int              `json:"rated_strength,omitzero"`
-	Quantity     fxp.Int              `json:"quantity"`
-	Level        fxp.Int              `json:"level,omitzero"`
-	Uses         int                  `json:"uses,omitzero"`
-	Equipped     bool                 `json:"equipped,omitzero"`
+	VTTNotes     string            `json:"vtt_notes,omitzero"`
+	Replacements map[string]string `json:"replacements,omitempty"`
+
+	// Leverage a default generic implementation of the Modifiable constraint that makes this a ModifiableNode
```
richardwilkes: There is no "ModifiableNode", so I assume this was meant to read "modifiable node" instead?

model/gurps/equipment_modifier_test.go
richardwilkes: Still refers to the old name ("setEquipment()")

model/gurps/trait_modifier_test.go
richardwilkes: Still refers to the old name, "setTrait()"

model/gurps/modifiers.go
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 1998-2026 by Richard A. Wilkes. All rights reserved.
+//
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, version 2.0. If a copy of the MPL was not distributed with
+// this file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// This Source Code Form is "Incompatible With Secondary Licenses", as
+// defined by the Mozilla Public License, version 2.0.
+
+package gurps
```
richardwilkes: A lot of the stuff in this file doesn't actually seem to be used anywhere. I assume this is prep for a follow-on PR?

model/gurps/modifiers.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// SetModifiers sets the list of modifiers
+func (m *modifiable[T, TN]) SetModifiers(all []T) {
+	m.Modifiers = all
+}
+
+// AddModifier adds a modifier to the list
+func (m *modifiable[T, TN]) AddModifier(mod T) {
+	m.Modifiers = append(m.Modifiers, mod)
```
richardwilkes: This doesn't set the target/owner, unlike what used to happen in all the places where modifiers were added. This will result in incorrect behavior.

model/gurps/modifiers.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// SetTargetNode calls SetTargetNode on each modifier.
+func (m *modifiable[T, TN]) SetTargetNode(target TN) {
```
richardwilkes: This method should be named with the word Modifier in it, as otherwise it isn't obvious that it is only affecting modifiers... which wouldn't be a problem if this method (and the others on this object) weren't promoted into the Equipment and Trait objects.

model/gurps/modifiers.go
richardwilkes: Also, this should probably be merged with `modifier.go`. I don't care which file is kept, but lets keep their contents in one file.

model/gurps/modifiers.go
```diff
@@ -0,0 +1,70 @@
...
+// This Source Code Form is "Incompatible With Secondary Licenses", as
+// defined by the Mozilla Public License, version 2.0.
+
+package gurps
+
+// assertModifiable is used at compile time to check a *constraint*
+func assertModifiable[N Modifiable[MN, TN], MN ModifierNode[MN, TN], TN Node[TN]]() {}
+
+// Modifiable represents an interface for an object that accepts modifier nodes
+type Modifiable[T ModifierNode[T, TN], TN Node[TN]] interface {
```
richardwilkes: Why T here, but MN above? Seems like we should be consistent throughout.

model/gurps/modifiers.go
```diff
@@ -0,0 +1,70 @@
...
+// assertModifierNode is used at compile time to check a *constraint*
+func assertModifierNode[T ModifierNode[T, TN], TN Node[TN]]() {}
+
+// ModifierNode is a narrowed constraint for a Node
+// The narrowed constraint enforces that TargetNode and SetTargetNode functions are satisfied.
+// The narrowed constraint enforces the GeneralModifier interface is satisfied.
+type ModifierNode[N Node[N], TN Node[TN]] interface {
+	Node[N]
+	TargetNode() TN
+	SetTargetNode(TN)
```
richardwilkes: I'd prefer something like `TargetOwner()` and `SetTargetOwner()` instead of the Node variants, as Node doesn't convey the same kind of ownership as implied here.

model/gurps/trait.go
```diff
@@ -93,10 +94,14 @@ type TraitEditData struct {
 	VTTNotes     string            `json:"vtt_notes,omitzero"`
 	UserDesc     string            `json:"userdesc,omitzero"`
 	Replacements map[string]string `json:"replacements,omitempty"`
-	Modifiers    []*TraitModifier  `json:"modifiers,omitempty"`
-	SelfControl  selfctrl.Roll     `json:"cr,omitzero"`
-	Frequency    frequency.Roll    `json:"frequency,omitzero"`
-	Disabled     bool              `json:"disabled,omitzero"`
+
+	// Leverage a default generic implementation of the Modifiable constraint that makes this a ModifiableNode
```
richardwilkes: Same comment as earlier: There is no "ModifiableNode", so I assume this was meant to read "modifiable node" instead?

richardwilkes (changes requested):

model/gurps/modifiers.go
LeeSaferite: Yes, some is prep work.

LeeSaferite: Merged.

model/gurps/modifiers.go
LeeSaferite: We settled on `Target` and `SetTarget`

model/gurps/modifiers.go
LeeSaferite: Cleaned up.

model/gurps/modifiers.go
LeeSaferite: The existing code is just use `append`, so it already has to set the target. There's no *simple* way to make it auto-set the target. I did make `SetTarget` return itself, so it can be used fluent.

Example
```
t := &Trait{}
tm := &TraitModifier{}
t.AddModifier(tm.SetTarget(t))
```

The only way to make it automatic is to have `AddModifier` live at the `*Trait` level.

PR #1131, opened 2026-09-10
Adding Headless API Run Mode

LeeSaferite:
This run mode is build flag gated and designed to be used during local
automated integration testing. It's provided so the application can run
in a "headless" manner where it's rendered into a frame buffer that is
exposed via a simple HTTP API. This same API also provides a path for
input as well as some additional endpoints to inspect the internal UI
structures.

NOTE: This *will not* be part of a normal build. The `headlessapi` build
flag is needed to include this feature.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -10,8 +10,11 @@
 package main
 
 import (
+	"flag"
+	"io"
 	"testing"
 
+	"github.com/richardwilkes/gcs/v5/runmode"
 	"github.com/richardwilkes/toolbox/v2/check"
 )
 
@@ -59,4 +62,44 @@ func TestExclusiveModeMsg(t *testing.T) {
 			c.Equal("", msg, tc.name)
 		}
 	}
+}
+
+// TestRunModeFactories checks the wiring every registered runmode.Factory is expected to provide: it registers a flag
+// named after the Mode it returns, that Mode is not requested until that flag is given, and it is requested once it
+// is. Each factory is called with its own flag set, so this exercises exactly what main does with flag.CommandLine
+// without disturbing it. Which modes take part depends on what got compiled into this build (see runmode.Factories).
+func TestRunModeFactories(t *testing.T) {
+	c := check.New(t)
+	c.NotEqual(0, len(runmode.Factories))
+	seen := make(map[string]bool, len(runmode.Factories))
+	for _, factory := range runmode.Factories {
+		flagSet := flag.NewFlagSet("gcs", flag.ContinueOnError)
+		flagSet.SetOutput(io.Discard)
+		m := factory(flagSet)
+		c.NotEqual("", m.Name)
+		c.False(seen[m.Name], m.Name)
+		seen[m.Name] = true
+		c.NotNil(m.Requested, m.Name)
+		c.NotNil(m.Start, m.Name)
+		f := flagSet.Lookup(m.Name)
+		c.NotNil(f, m.Name)
+		if f == nil {
+			continue
+		}
+		for _, hidden := range m.HiddenFlagNames {
+			c.NotNil(flagSet.Lookup(hidden), m.Name, hidden)
+		}
+		c.False(m.Requested(), m.Name)
+		c.NoError(flagSet.Parse([]string{"-" + m.Name + "=" + requestValueFor(f)}), m.Name)
+		c.True(m.Requested(), m.Name)
+	}
+}
+
+// requestValueFor returns a value for f that asks for the run mode it belongs to: bool flags are simply turned on,
+// while the rest -- the ones that name a file or an address -- are considered requested when they are non-empty.
+func requestValueFor(f *flag.Flag) string {
+	if b, ok := f.Value.(interface{ IsBoolFlag() bool }); ok && b.IsBoolFlag() {
+		return "true"
+	}
+	return "requested"
 }
--- a/main_test.go
+++ b/main_test.go
@@ -18,28 +18,18 @@ import (
 func TestExclusiveModeMsg(t *testing.T) {
 	c := check.New(t)
 	for _, tc := range []struct {
-		name             string
-		textTmplPath     string
-		finishUpdatePath string
-		wantContains     []string
-		convert          bool
-		sync             bool
-		wantErr          bool
+		name         string
+		modes        []string
+		wantContains []string
+		wantErr      bool
 	}{
 		{name: "none specified"},
-		{name: "convert only", convert: true},
-		{name: "sync only", sync: true},
-		{name: "text only", textTmplPath: "tmpl"},
-		{name: "finish-update only", finishUpdatePath: "state.json"},
-		{name: "convert and sync", convert: true, sync: true, wantErr: true, wantContains: []string{"--convert", "--sync"}},
-		{name: "convert and text", convert: true, textTmplPath: "tmpl", wantErr: true, wantContains: []string{"--convert", "--text"}},
-		{name: "sync and text", sync: true, textTmplPath: "tmpl", wantErr: true, wantContains: []string{"--sync", "--text"}},
-		{name: "finish-update and convert", convert: true, finishUpdatePath: "state.json", wantErr: true, wantContains: []string{"--convert", "--finish-update"}},
-		{name: "finish-update and text", textTmplPath: "tmpl", finishUpdatePath: "state.json", wantErr: true, wantContains: []string{"--text", "--finish-update"}},
-		{name: "all three", convert: true, sync: true, textTmplPath: "tmpl", wantErr: true, wantContains: []string{"--convert", "--sync", "--text"}},
-		{name: "all four", convert: true, sync: true, textTmplPath: "tmpl", finishUpdatePath: "state.json", wantErr: true, wantContains: []string{"--convert", "--sync", "--text", "--finish-update"}},
+		{name: "one mode", modes: []string{"--convert"}},
+		{name: "two modes", modes: []string{"--convert", "--sync"}, wantErr: true, wantContains: []string{"--convert", "--sync"}},
+		{name: "four modes", modes: []string{"--convert", "--sync", "--text", "--finish-update"}, wantErr: true, wantContains: []string{"--convert", "--sync", "--text", "--finish-update"}},
+		{name: "five modes", modes: []string{"--convert", "--sync", "--text", "--finish-update", "--headless-api"}, wantErr: true, wantContains: []string{"--convert", "--sync", "--text", "--finish-update", "--headless-api"}},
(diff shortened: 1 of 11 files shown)


ux/headless_api.go
```diff
@@ -0,0 +1,620 @@
...
+		http.Error(w, "unknown op: "+req.Op, http.StatusBadRequest)
+		return
+	}
+	s.drainSessionErrors()
+	writeJSON(w, struct {
+		OK bool `json:"ok"`
+	}{OK: true})
+}
+
+func parseMods(names []string) mod.Modifiers {
```
richardwilkes, line 363: Things like this, and the key code parsing below, should probably be done using the existing mechanisms (like mod.FromKey()) so that future changes are automatically accounted for.

ux/headless_api.go
```diff
@@ -0,0 +1,620 @@
...
+	case "right":
+		return unison.ButtonRight
+	case "middle":
+		return unison.ButtonMiddle
+	default:
+		return unison.ButtonLeft
+	}
+}
+
+var namedKeys = map[string]unison.KeyCode{
```
richardwilkes: As noted earlier, you should probably be using unison.KeyCodeFromKey() to turn text into the equivalent key codes.

ux/headless_api.go
```diff
@@ -0,0 +1,620 @@
...
+	var result inspectResult
+	var ok bool
+	s.screen.Do(func() {
+		wnd := s.screen.WindowAt(pt)
+		if wnd == nil {
+			return
+		}
+		ok = true
+		local := pt.Sub(wnd.ContentRect().Point)
+		result = describeWindowAndPanel(wnd, wnd.Content().PanelAt(local))
```
richardwilkes: From claude:

/inspect hit-tests in the wrong coordinate space (ux/headless_api.go:400). It calls wnd.Content().PanelAt(local) with root-space coordinates, but headless uses the in-window menu bar, which offsets the content panel by the bar height. Every lookup lands one bar height too low, and menu titles, open menu popups, and tooltips (all root children) can never be found. Fix: hit-test from the root as Window.mouseDown does, e.g. wnd.Content().Parent().PanelAt(local), matching the existing headless test helpers.

ux/headless_api.go
```diff
@@ -0,0 +1,620 @@
...
+	// The workspace's own settings-saving close handler (workspaceWillClose, in workspace.go) only ever runs when the
+	// main window is closed the normal way, and nothing here ever closes it -- the process just runs until killed.
+	// Without this, everything settings-backed (page reference mappings, recent files, window layout, and so on) that
+	// a session here changes is silently lost the moment the container stops, however it stops. A periodic autosave
+	// covers an unclean stop (SIGKILL, OOM, a crash); the signal handler below covers a clean one (SIGINT/SIGTERM,
+	// which is what "podman stop"/"docker stop" send, and Ctrl-C) by saving once more and only then letting the
+	// process exit, so the common case loses nothing at all.
+	stopAutosave := make(chan struct{})
+	go autosaveSettings(screen, stopAutosave)
+	sigCh := make(chan os.Signal, 1)
```
richardwilkes: Rather than rolling your own, use xos.RunAtExit()

ux/headless_api.go
```diff
@@ -0,0 +1,620 @@
...
+		saveSettings(screen)
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if shutdownErr := httpServer.Shutdown(shutdownCtx); shutdownErr != nil {
+			errs.Log(shutdownErr)
+		}
+	}()
+
+	if err = httpServer.Serve(listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		errs.Log(err)
```
richardwilkes: Should probably call xos.Exit(1) or something similar here, to indicate things didn't go as planned

main_test.go
```diff
@@ -18,28 +18,18 @@ import (
...
-		{name: "finish-update and text", textTmplPath: "tmpl", finishUpdatePath: "state.json", wantErr: true, wantContains: []string{"--text", "--finish-update"}},
-		{name: "all three", convert: true, sync: true, textTmplPath: "tmpl", wantErr: true, wantContains: []string{"--convert", "--sync", "--text"}},
-		{name: "all four", convert: true, sync: true, textTmplPath: "tmpl", finishUpdatePath: "state.json", wantErr: true, wantContains: []string{"--convert", "--sync", "--text", "--finish-update"}},
+		{name: "one mode", modes: []string{"--convert"}},
+		{name: "two modes", modes: []string{"--convert", "--sync"}, wantErr: true, wantContains: []string{"--convert", "--sync"}},
+		{name: "four modes", modes: []string{"--convert", "--sync", "--text", "--finish-update"}, wantErr: true, wantContains: []string{"--convert", "--sync", "--text", "--finish-update"}},
+		{name: "five modes", modes: []string{"--convert", "--sync", "--text", "--finish-update", "--headless-api"}, wantErr: true, wantContains: []string{"--convert", "--sync", "--text", "--finish-update", "--headless-api"}},
```
richardwilkes: From claude:

TestExclusiveModeMsg can't catch a format regression (main_test.go:31). Fixtures pass already-dashed names (--convert) while production passes bare names, so dropping the - prefix from the format string would still pass. Use bare names and assert on -convert / -sync.

ux/headless_api.go
```diff
@@ -0,0 +1,620 @@
...
+	screen.Do(func() {
+		if err := gurps.GlobalSettings().Save(); err != nil {
+			errs.Log(err)
+		}
+	})
+}
+
+type headlessAPIServer struct {
+	screen *unison.HeadlessScreen
+	wnd    *unison.Window
```
richardwilkes: From claude:

headlessAPIServer.wnd is dead state (ux/headless_api.go:194). Written in the startup callback, never read. Use a local variable and drop the field.

Lets get rid of the empty line, too.

richardwilkes (changes requested):

ux/headless_api.go
```diff
@@ -0,0 +1,673 @@
...
+func (s *headlessAPIServer) handleInspectFocus(w http.ResponseWriter, _ *http.Request) {
+	var result inspectResult
+	var ok bool
+	s.screen.Do(func() {
+		wnd := s.screen.FocusedWindow()
+		if wnd == nil {
+			return
+		}
+		ok = true
+		result = describeWindowAndPanel(wnd, wnd.Focus())
```
richardwilkes: Calling wnd.Focus() will actually set a focus if one isn't already present, so it technically mutates the state, which the "inspect" prefix seems to indicate it shouldn't be doing. Not sure if this is a real problem for your use cases, though.

ux/headless_api.go
```diff
@@ -0,0 +1,673 @@
...
+		close(stopAutosave)
+		saveSettings(screen)
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if shutdownErr := httpServer.Shutdown(shutdownCtx); shutdownErr != nil {
+			errs.Log(shutdownErr)
+		}
+	})
+
+	if err = httpServer.Serve(listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
```
richardwilkes: The server could be left running even after a user quit the program. You need to listen in on the headless screen's Running() or Done() methods.

ux/headless_api.go
```diff
@@ -0,0 +1,673 @@
...
+	d := describePanel(leaf, origin)
+	result.Panel = &d
+	for anc := leaf.Parent(); anc != nil; anc = anc.Parent() {
+		result.Ancestors = append(result.Ancestors, describePanel(anc, origin))
+	}
+	return result
+}
+
+func describePanel(p *unison.Panel, origin geom.Point) panelDTO {
+	rect := p.RectToRoot(p.ContentRect(false))
```
richardwilkes, line 514: You may want an option that clips this to the visible portion of the panel, since if the panel is inside a scroll panel, it could be significantly larger than the actual visible area, or could be offscreen entirely.

This could be particularly problematic for targeting mouse input, as if you aim it at a section that is in fact clipped, you'll be clicking on something you didn't intend to.

ux/headless_api.go
```diff
@@ -0,0 +1,673 @@
...
+	docFormatJSON
+)
+
+// negotiateDocFormat picks a docFormat from the value of an Accept header: docFormatJSON for application/json,
+// docFormatYAML for application/yaml or application/x-yaml, and docFormatMD -- the default -- for anything else,
+// including an empty or missing header, "*/*", or text/markdown itself. Ties (equal, and equal to the highest,
+// "q" value) are broken by whichever the header lists first.
+func negotiateDocFormat(accept string) docFormat {
+	best := docFormatMD
+	bestQ := -1.0
```
richardwilkes: Maybe this should start at 0 so that clients that can't handle a particular format (i.e. they report 0 for q) don't inadvertently get that format.

ux/headless_api.yaml
```diff
@@ -0,0 +1,314 @@
...
+                modifiers:
+                    type: array
+                    items:
+                        type: string
+        Rect:
+            type: object
+            properties:
+                x:
+                    type: number
+                    required: true
```
richardwilkes: From claude:

The served OpenAPI 3.0.3 document is invalid: 'required: true' is used inside property schemas (Rect.x/y/w/h, lines 249-258) instead of an object-level 'required' array, and 'required: true' sits as a sibling of $ref on the /inspect x and y parameters (lines 80/82), which 3.0 ignores.

main_test.go
```diff
@@ -18,28 +18,38 @@ import (
...
+		},
+		{
+			name:         "five modes",
+			modes:        []string{"convert", "sync", "text", "finish-update", "headless-api"},
+			wantErr:      true,
+			wantContains: []string{"-convert", "-sync", "-text", "-finish-update", "-headless-api"},
+		},
 	} {
-		msg := exclusiveModeMsg(tc.convert, tc.sync, tc.textTmplPath, tc.finishUpdatePath)
+		msg := exclusiveModeMsg(tc.modes)
```
richardwilkes, line 55: From claude:

The Factory's *flag.FlagSet parameter is never passed non-nil, its nil->flag.CommandLine fallback is copy-pasted into all five factories (also lines 47, 66, updater/run_modes.go:32, ux/headless_api.go:143), and no test uses it, so the flag-to-mode wiring the old TestExclusiveModeMsg exercised is now untested.

richardwilkes (changes requested):

richardwilkes (approved):
Will address the startup issue separately

PR #1134, opened 2026-09-14
Keep template choices from landing where they can't be managed

LeeSaferite:
The "Choices" a container carries are only editable on a template, but until now such a container could be dragged or copied onto a character sheet or a library list, where nothing offers a way to manage them once they are there.

Copying or dropping one onto a character sheet now makes those choices at once, as a partial template application. The same dialogs an applied template presents are shown, and each container carrying choices is replaced by what was picked. Canceling leaves the destination as it was.

Anywhere else the copy warns that the choices cannot be kept there. Accepting from the warning dialog copies the rows with the choices removed and everything else about them untouched. Declining copies nothing.


ux/template_picker_constraint.go
```diff
@@ -0,0 +1,232 @@
...
+// decideTemplatePickerCopy reports what a copy of rows carrying template choices onto the given destination is to do
+// about them. Only a template takes them as they are, since nothing else provides a way to manage them once they have
+// arrived. A character sheet makes the choices on the spot, which is what applying a template does. Anywhere else --
+// a library list, most of all -- can only take the rows without the choices, which is confirmed here before it does.
+func decideTemplatePickerCopy(dst unison.Paneler) templatePickerCopyAction {
+	switch unison.AncestorOrSelf[unison.Dockable](dst).(type) {
+	case *Template:
+		return templatePickerCopyKeep
+	case *Sheet:
+		return templatePickerCopyResolve
```
richardwilkes, line 92: LootSheets should be here, too.

ux/table_drop.go
```diff
@@ -259,6 +274,35 @@ func willDropCallback[T gurps.Node[T]](from, to *unison.Table[*Node[T]], move bo
...
+	case templatePickerCopyResolve:
+		// The drop was let through on the condition that the template choices it carries are made now, turning it into
+		// a partial template application.
+		pendingTemplatePickerAction = templatePickerCopyKeep
+		// unison doesn't erase the drop feedback until the drop callback this was reached from returns, which is well
+		// after the choice dialogs below have come and gone, so it is erased here rather than left behind them.
+		if to.DragExitedCallback != nil {
+			to.DragExitedCallback()
+		}
+		if !resolveTemplatePickers(to, to.SelectedRows(true), processPickerRows) {
```
richardwilkes, line 289: By using `to.SelectedRows(true)`, you are asking it to prune out undisclosed nodes and nodes that aren't part of the immediate selection. If the drop is into an undisclosed node, it effectively resolves to nothing, bypassing the template check.

ux/table.go
```diff
@@ -378,17 +378,42 @@ func copySelectionTo[T gurps.Node[T], D copyDestination](table *unison.Table[*No
 	}
 	sel := table.SelectedRows(true)
 	key := blockKeyForRow(sel[0].Data())
+	// Template choices can only be managed on a template, so copying rows that carry them onto a character sheet turns
+	// into a partial template application instead. Every destination of a single copy is of the same kind, so what to
+	// do about them is decided once rather than once per destination.
+	pickerAction := templatePickerCopyKeep
+	if rowsHaveTemplatePickers(ExtractNodeDataFromList(sel)) {
+		if pickerAction = decideTemplatePickerCopy(destinations[0]); pickerAction == templatePickerCopyRefuse {
```
richardwilkes, line 386: The pickerAction will never be refuse in this case, as only sheets & templates can be destinations. The drag & drop path doesn't go through this, so the case the refuse action is for won't ever fire... safe enough to leave, but wanted to make sure that was in fact the intent.

ux/table_drop.go
```diff
@@ -127,6 +127,21 @@ func modifierAltDropSupport[T gurps.Node[T], M gurps.Node[M]](p *listProvider[T]
...
 	table.ClientData()[TableProviderClientKey] = provider
 	table.InstallDropSupport(provider.DragKey(), provider.DropShouldMoveData, willDropCallback[T], didDropCallback[T])
+
+	// Template choices can only be managed on a template, so a drop that would carry them anywhere else is handled
+	// here, before unison has inserted anything, rather than being unwound afterwards. The guard also decides whether
+	// a drop onto a sheet is to resolve those choices instead (see guardTemplatePickerDrop).
+	dropCallback := table.DropCallback
+	table.DropCallback = func(di drag.Info, where geom.Point, mods mod.Modifiers) bool {
+		if !guardTemplatePickerDrop(table, provider, di) {
+			if table.DragExitedCallback != nil {
```
richardwilkes, line 137: Seems like this should be done before the potential dialogs are shown, for the same reason lines 284-288 note.

ux/template_picker.go
```diff
@@ -41,6 +41,11 @@ func processPickerRows[T gurps.Node[T]](rows []*Node[T]) (revised []*Node[T], ab
 	return revised, false
 }
 
+// templatePickerChoiceHook, when non-nil, stands in for the template choice dialog, answering with the indexes of the
+// children to keep or asking for the operation to be canceled. Headless tests set this, since they have no way to
+// respond to a modal dialog.
+var templatePickerChoiceHook func(row any, childCount int) (keep []int, cancel bool)
```
richardwilkes, line 47: From claude:

**ux/template_picker.go:47,74** — Medium — `templatePickerChoiceHook` is a test-only seam added to production code,
  against the repo's established pattern for this exact need, which is parameter injection
  (`applyTemplateToSheetWithPickers(sheet, suppress, processPickers)`, `ux/template.go:249`). It also short-circuits
  `processPickerRow` after the `IsZero` early-out but before the checkbox and `tp.Qualifier.Matches(total)` machinery,
  so tests can assert outcomes the real dialog cannot produce (keeping two children under a "pick exactly 1"
  qualifier). Meanwhile the `process` parameter the PR adds to `resolveTemplatePickers`
  (`ux/template_picker_constraint.go:153`) is documented as the substitution point for headless tests, but every
  caller, production and test alike, passes `processPickerRows`; the tests steer behavior through the global hook
  instead. One mechanism should remain, preferably the injected one.

ux/template_picker_constraint.go
```diff
@@ -0,0 +1,232 @@
...
+		if rowsHaveTemplatePickers(row.NodeChildren()) {
+			return true
+		}
+	}
+	return false
+}
+
+// stripTemplatePickers removes the template choices from the rows and everything beneath them, leaving them otherwise
+// untouched.
+func stripTemplatePickers[T gurps.Node[T]](rows []T) {
```
richardwilkes, line 69: From claude:

**ux/template_picker_constraint.go:69-78** — Medium — Stripping the choices is not durable. `TemplatePicker` lives in
  `TraitContainerSyncData` / `SkillContainerOnlySyncData` / `SpellContainerOnlySyncData`, which is source-sync data:
  it is hashed (`model/gurps/trait.go:1059`) and `SyncWithSource` copies the whole struct back from the library
  original (`model/gurps/trait.go:1017-1030`, likewise skill and spell). Because the strip deliberately keeps the
  row's `Source`, a stripped container immediately shows as out of sync with its library source, and "Sync with
  Source" (`SyncWithSourceForSelection`, `ux/table.go:578`) re-installs the template choices in the very list where
  they cannot be managed. Clearing the `Source` when stripping, or refusing to strip a sourced row, would be
  consistent.

richardwilkes (changes requested):

PR #1136, opened 2026-09-14
Adding point range support

LeeSaferite:
The current system has no ability to display when and item has a point
range and not a single point value. This is most common on templates
where a choice container has a range of point valued items. Until the
choice is made, the exact points are unknown. It can also appear from
library items when an item has required selections before a final point
cost is known.


Xain3:
I gave it a first run and it works better than I ever expected! The visualisation of the range seem to work perfectly, I haven''t found a way to break it and I like how it supports all options... and skills too!

model/gurps/skill.go
```diff
@@ -644,9 +644,20 @@ func (s *Skill) SetRawPoints(points fxp.Int) bool {
...
+// costs the same reports that cost; see PointsRange for one whose outcomes differ.
 func (s *Skill) AdjustedPoints(tooltip *xbytes.InsertBuffer) fxp.Int {
 	if s.Container() {
+		// A container that presents a choice is never worth what its children add up to, since only some of them will
+		// be taken. When every way of making that choice costs the same -- "pick 20 points worth", most often -- that
+		// is what it is worth. When they don't, there is no single answer, and the total of the children is left as
+		// the answer it has always given here, with PointsRange holding the one that can be relied upon. The range is
+		// asked for without the tooltip, since the notes it would gather describe children that may not be taken.
+		if !s.TemplatePicker.IsZero() {
+			if value, settled := s.PointsRange(nil).Settled(); settled {
```
LeeSaferite, line 657: This should probably pass the `tooltip`.

model/gurps/spell.go
```diff
@@ -902,9 +902,20 @@ func (s *Spell) SetRawPoints(points fxp.Int) bool {
...
+// costs the same reports that cost; see PointsRange for one whose outcomes differ.
 func (s *Spell) AdjustedPoints(tooltip *xbytes.InsertBuffer) fxp.Int {
 	if s.Container() {
+		// A container that presents a choice is never worth what its children add up to, since only some of them will
+		// be taken. When every way of making that choice costs the same -- "pick 20 points worth", most often -- that
+		// is what it is worth. When they don't, there is no single answer, and the total of the children is left as
+		// the answer it has always given here, with PointsRange holding the one that can be relied upon. The range is
+		// asked for without the tooltip, since the notes it would gather describe children that may not be taken.
+		if !s.TemplatePicker.IsZero() {
+			if value, settled := s.PointsRange(nil).Settled(); settled {
```
LeeSaferite, line 915: This should probably pass the `tooltip`.

model/gurps/trait.go
```diff
@@ -583,33 +586,113 @@ func (t *Trait) ResolvedMaxLevels() fxp.Int {
...
+		return PointsRangeOf(0)
+	}
+	if value, settled := settledPickerCost(t.TemplatePicker); settled {
+		return PointsRangeOf(value)
+	}
+	if !t.TemplatePicker.IsZero() {
+		// The notes the children would contribute describe items that may not be taken, so they are left out.
+		ranges := make([]PointsRange, len(t.Children))
+		for i, one := range t.Children {
+			ranges[i] = one.PointsRange(nil)
```
LeeSaferite, line 642: This should probably pass the `tooltip`.

model/gurps/trait.go
```diff
@@ -583,33 +586,113 @@ func (t *Trait) ResolvedMaxLevels() fxp.Int {
...
 		return AdjustedPoints(EntityFromNode(t), t, t.CanLevel, t.BasePoints, t.Levels, t.PointsPerLevel,
 			t.SelfControl, t.Frequency, t.AllModifiers(), t.RoundCostDown)
 	}
+	// A container that presents a choice is never worth what its children add up to, since only some of them will be
+	// taken. When every way of making that choice costs the same -- "pick 20 points worth", most often -- that is what
+	// it is worth. When they don't, there is no single answer, and the total of the children is left as the answer it
+	// has always given here, with PointsRange holding the one that can be relied upon. The range is asked for without
+	// the tooltip, since the notes it would gather describe children that may not be taken.
+	if !t.TemplatePicker.IsZero() {
+		if value, settled := t.PointsRange(nil).Settled(); settled {
```
richardwilkes, line 605: Another place where tooltip should probably be passed in

model/gurps/trait.go
richardwilkes: Hmm... this is a case unlike the others, where things were settled. If it is added here, there are a number of other places above (in the GitHub diff view, anyway) that should also be handed the tooltip

model/gurps/points_range.go
```diff
@@ -0,0 +1,420 @@
...
+			// Children that between them cost less than the picker asks for set no ceiling on it. That isn't a
+			// malformed template: the children of such a picker are typically skills or spells carrying no points at
+			// all, whose cost is assigned while picking (see ux.pickerRowPointEditor), and a leveled trait's cost can
+			// be raised there too. With nothing to bound it, the container is worth what it asks for or more.
+			if !upper.unlimited && upper.total < qualifier {
+				return pointsRangeAtLeast(qualifier)
+			}
+			return PointsRange{Min: &qualifier, Max: upper.end()}
+		case criteria.AtMostNumber:
+			upper := everythingPositive(children)
```
richardwilkes, line 249: This logic seems to assume the "at most" number will always be positive, while there is nothing preventing a choice from being "at most -20".

model/gurps/points_range.go
```diff
@@ -0,0 +1,420 @@
...
+				Max: everythingPositive(children).end(),
+			}
+		}
+	case picker.Points:
+		qualifier := tp.Qualifier.Qualifier
+		switch compare {
+		case criteria.EqualsNumber:
+			return PointsRangeOf(qualifier)
+		case criteria.AtLeastNumber:
+			upper := everythingPositive(children)
```
richardwilkes, line 239: This logic seems to assume the "at least" number will always be positive, while there is nothing preventing a choice from being "at least -20".

richardwilkes (changes requested):

model/gurps/trait.go
```diff
@@ -584,33 +587,114 @@ func (t *Trait) ResolvedMaxLevels() fxp.Int {
...
 		return AdjustedPoints(EntityFromNode(t), t, t.CanLevel, t.BasePoints, t.Levels, t.PointsPerLevel,
 			t.SelfControl, t.Frequency, t.AllModifiers(), t.RoundCostDown)
 	}
+	// A container that presents a choice is never worth what its children add up to, since only some of them will be
+	// taken. When every way of making that choice costs the same -- "pick 20 points worth", most often -- that is what
+	// it is worth. When they don't, there is no single answer, and the total of the children is left as the answer it
+	// has always given here, with PointsRange holding the one that can be relied upon. The tooltip is handed along, but
+	// a container never puts anything into it; see PointsRange for why.
+	if !t.TemplatePicker.IsZero() {
+		if value, settled := t.PointsRange(tooltip).Settled(); settled {
```
richardwilkes: This will show the wrong value for Alternative Abilities containers, which do special pricing depending on the number of slots which doesn't seem to be accounted for here. Not sure if it is worth trying to make that work, but wanted to point out the discrepancy.

ux/template_picker.go
```diff
@@ -113,19 +113,21 @@ func processPickerRow[T gurps.Node[T]](row T) (revised []T, abort bool) {
...
-					total += fxp.One
+					total = total.Add(gurps.PointsRangeOf(fxp.One))
 				case picker.Points:
-					total += rawPoints(children[i])
+					total = total.Add(pointsRangeFor(children[i]))
 				}
 			}
 		}
-		matches := tp.Qualifier.Matches(total)
+		matches := total.CanSatisfy(tp.Qualifier)
```
richardwilkes, line 130: From claude:

**ux/template_picker.go:130** — `CanSatisfy` enables OK when *some* completion of a checked nested picker could satisfy the parent, but nothing re-checks the parent's constraint after the nested dialogs resolve, so the final selection can silently violate the parent picker without going through Override. Parent "pick exactly 20 points"; the user checks a child container "pick at least 5 points" over {5, 30} (range `5+`). `CanSatisfy(Equals 20)` is true so OK enables; in the nested dialog (line 227) the user picks the 30-point item. Result: 30 points applied under an "exactly 20" picker with no warning, whereas an unsettled leaf total would have disabled OK. The comment on `pointsRangeForPicker` acknowledges achievability isn't checked, but at the point where the user commits the pick it produces a wrong answer. Resolve nested pickers first, or re-check the outer qualifier after they resolve, before accepting the pick; at minimum mark the progress total as provisional.

model/gurps/cell.go
```diff
@@ -98,6 +98,23 @@ func fillTagsCell(data *CellData, tags []string) {
...
+	data.Type = cell.Text
+	data.Primary = r.String()
+	data.Alignment = align.End
+	if r.IsSettled() {
+		return
+	}
+	if data.Tooltip == "" {
+		data.Tooltip = PointsRangeTooltip()
+	} else {
+		data.Tooltip = PointsRangeTooltip() + "\n---\n" + data.Tooltip
```
richardwilkes, line 114: From claude:

**model/gurps/cell.go:114** — The "keep the caller's tooltip alongside" branch in `fillPointsCell` is unreachable: only a container can be unsettled, and containers never fill the tooltip, so `data.Tooltip` is always empty when the range is open. The `"\n---\n"` concatenation can never execute, yet the doc comment describes it as a supported state. Drop the branch or make it reachable.

model/gurps/points_range.go
```diff
@@ -0,0 +1,455 @@
...
+	case criteria.AtLeastNumber: // The pick costs the qualifier or more.
+		if costliest != nil && *costliest < cq.Qualifier {
+			return pointsRangeAtLeast(cq.Qualifier) // The children set no ceiling on what the picker asks for.
+		}
+		minimum := cq.Qualifier
+		if cheapest != nil && *cheapest > minimum {
+			minimum = *cheapest // The qualifier asks for less than the cheapest pick already costs.
+		}
+		return PointsRange{Min: &minimum, Max: costliest}
+	case criteria.AtMostNumber: // The pick costs the qualifier or less.
```
richardwilkes: From claude:

**model/gurps/points_range.go:286** — `pointsRangeForPickerByPoints` treats "at most N" and "at least N" asymmetrically over zero-point children. For AtLeast the ceiling is left open (`N+`) with the stated reason that skill and spell cost is assigned while picking. For AtMost the ceiling is clamped to `costliest`, which is 0 for zero-point skills, so a skill template container "pick at most 10 points" over three 0-point skills reports a settled `0`: the Pts column shows `0` instead of `≤10`, `AdjustedPoints` returns 0, and inside a parent "pick 20 points" picker the row contributes exactly 0 so OK stays disabled even though picking it assigns points. The tests only cover the negative-qualifier variant of this branch. Apply the same "the qualifier binds its end" rule to AtMost, or clamp AtLeast the same way, and add a zero-point-children test for both.

model/gurps/points_range.go
```diff
@@ -0,0 +1,455 @@
...
+// worth, given the ranges of the children that may be picked from.
+//
+// The count cases are exact: the cheapest way to satisfy "pick 3" is the 3 cheapest children, and a child that costs
+// less than nothing is always worth taking when the picker allows more to be taken. The points cases lean on the fact
+// that a points picker measures the very quantity it constrains -- the cost of what is picked -- so the qualifier
+// bounds the container's total directly, with no need to search for a subset that adds up to it. The cost of that
+// shortcut is that achievability isn't checked: "pick at least 7 points" from children worth 5 and 10 reports a
+// minimum of 7, where the cheapest satisfying pick is really 10. Reporting the constraint the picker states is both
+// cheaper and closer to how the picker describes itself.
+func pointsRangeForPicker(tp TemplatePicker, children []PointsRange) PointsRange {
```
richardwilkes, line 277: From claude:

**model/gurps/points_range.go:188** — `pointsRangeForPicker` short-circuits an empty child list to a settled `0` for every picker type. That contradicts `settledPickerCost`, which every `PointsRange` method consults first and which returns the qualifier for a Points/Equals picker regardless of children, and it contradicts the open-ended answers the points branch gives once a single child exists: a template container "pick at least 10 points" with no children shows `0`, then flips to `10+` when the first 0-point skill is added. A visible consequence of the `settledPickerCost` side is that an empty "pick 20 points" container now contributes 20 to a template's total where it previously contributed 0. Decide which answer is right for the empty case, make both paths agree, and pin it in `TestPointsRangeForExactPointsPickerDirectly`.

model/gurps/points_range.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+// pointsSortValue returns the lower end of a rendered point cost, along with whether that end has no limit, in which
+// case nothing finite sorts ahead of it.
+func pointsSortValue(text string) (value fxp.Int, unlimited bool) {
+	text = strings.TrimSpace(text)
+	if strings.HasPrefix(text, unboundedMinPrefix) || text == noLimitsAtAll {
+		return 0, true
+	}
+	value, _ = fxp.Extract(strings.ReplaceAll(text, ",", ""))
```
richardwilkes: There is no need to strip the commas from the text -- fxp.Extract() already handles that.

model/gurps/points_range.go
```diff
@@ -0,0 +1,455 @@
...
+		case criteria.AtLeastNumber:
+			compare = criteria.AnyNumber
+		}
+	}
+
+	cheapestFirst := byMin(children)
+	costliestFirst := byMax(children)
+
+	switch compare {
+	case criteria.EqualsNumber: // Expects and exact number of selections
```
richardwilkes: "Expects and exact" -> "Expects an exact"

model/gurps/trait.go
```diff
@@ -584,33 +587,114 @@ func (t *Trait) ResolvedMaxLevels() fxp.Int {
...
 	if !t.Container() {
 		return AdjustedPoints(EntityFromNode(t), t, t.CanLevel, t.BasePoints, t.Levels, t.PointsPerLevel,
 			t.SelfControl, t.Frequency, t.AllModifiers(), t.RoundCostDown)
 	}
+	// A container that presents a choice is never worth what its children add up to, since only some of them will be
+	// taken. When every way of making that choice costs the same -- "pick 20 points worth", most often -- that is what
+	// it is worth. When they don't, there is no single answer, and the total of the children is left as the answer it
+	// has always given here, with PointsRange holding the one that can be relied upon. The tooltip is handed along, but
+	// a container never puts anything into it; see PointsRange for why.
+	if !t.TemplatePicker.IsZero() {
```
richardwilkes, line 602: From claude:

**model/gurps/trait.go:605** — For a container with an unsettled picker, `AdjustedPoints` walks the subtree twice: once via `PointsRange`, which builds every child's range, and again in the fallback sum. Each nested unsettled container repeats this, so the cost grows with nesting depth, plus a `[]PointsRange` allocation per container per call. `AdjustedPoints` runs from `CellData` on every render and every sort comparison, and from filters and scripts. Computing the children's ranges once and deriving both the range and the plain sum from them avoids the second walk. Same shape in `skill.go:650` and `spell.go:908`.

model/gurps/trait.go
```diff
@@ -584,33 +587,114 @@ func (t *Trait) ResolvedMaxLevels() fxp.Int {
...
+		// The disabled case is covered too: AdjustedPoints reports nothing for a trait that is switched off.
+		return PointsRangeOf(t.AdjustedPoints(tooltip))
+	}
+	if t.EffectivelyDisabled() {
+		return PointsRangeOf(0)
+	}
+	if value, settled := settledPickerCost(t.TemplatePicker); settled {
+		return PointsRangeOf(value)
+	}
+	if !t.TemplatePicker.IsZero() {
```
richardwilkes: From claude:

**model/gurps/trait.go:641** — `PointsRange` in Trait, Skill and Spell each build the children's ranges in two separate branches (picker vs. none) with a duplicated `ranges := make(...)` loop (`trait.go:642/648`, `skill.go:698/704`, `spell.go:943/949`), even though `pointsRangeForPicker` already returns `sumPointsRanges` for a NotApplicable picker. The whole container-range method and the five-line `AdjustedPoints` comment are copied three times. One loop plus one `pointsRangeForPicker(tp, ranges)` call per method, or a generic `containerPointsRange[T Node[T]]` helper, removes the triplication.

richardwilkes (changes requested):
Lets also add two entries into the release notes file, one under new & improved for the range display, and another under bug fixes for the sorting bug fix when "group containers on sort" is enabled. Don't worry to much about the wording, just need enough to remind me what it is when I go through and prepare the actual release notes.

model/gurps/trait.go
LeeSaferite: The call to PointRange *does* handle AA containers. The only case where it won't is if the AA container is **also** a template picker container. That is an unsupported configuration that will soon be disallowed via code anyway. If someone does it today, the AA container will simply dissolve once the template is applied.

model/gurps/cell.go
LeeSaferite: I 100% disagree here. This code is written to support what is **possible**. A node is **allowed** to have a tooltip, so we need to handle that situation. The fact that our current container nodes aren't expected to surface a tooltip is just an artifact of what we do today. We could just as easily add a tooltip about some important detail tomorrow and then spend way too long tracing why the tooltip is lost. This branch is insurance against a future state where a tooltip **is** returned by a container.

richardwilkes: Fair enough

ux/template_picker.go
LeeSaferite: This is ok for the moment. Upcoming template changes will resolve this.

richardwilkes (approved):

PR #1144, opened 2026-09-19
General Housekeeping

LeeSaferite:
- Add `TemplatePickerNode` constraint
- Add `assertTemplatePickerNode` compile-time constraint assertion
- Add `HasTemplatePickerData` helper to check a node and it's descendants for template picker data
- Add `ClearTemplatePickerData` helper to clear template picker data from a node and it's descendants
- Use `assertTemplatePickerNode` on `*Trait`, `*Skill`, and `*Spell`
- Move some code to "proper" locations

This PR doesn't include any behavioral changes.


model/gurps/template_picker.go
```diff
@@ -26,6 +26,15 @@ type TemplatePickerProvider interface {
 	TemplatePickerData() ([]picker.Type, *TemplatePicker)
 }
 
+// TemplatePickerNode is a constraint for a Node that implements the TemplatePickerProvider interface
+type TemplatePickerNode[T TemplatePickerNode[T]] interface {
```
richardwilkes, line 30: This, and the two functions that use it below, seem to be dead code. I assume you plan to use them in some future PR?

Also, from claude:

`model/gurps/template_picker.go:29-36`: the `TemplatePickerNode` constraint is used by nothing but its own `assertTemplatePickerNode`. No generic function is constrained by it, and the two new helpers deliberately take `Node[T]` and type-assert every node at runtime instead, even though the result is constant for a given `T` because `NodeChildren` returns `[]T` (so calling them on an `*Equipment` or `*Note` tree walks every node doing nothing). The three new assertions in `trait.go:46`, `skill.go:42` and `spell.go:41` also re-verify exactly what the adjacent `assertNode[*X]` plus the existing `_ TemplatePickerProvider = &Trait{}` (`trait.go:50`, `skill.go:47`, `spell.go:46`) already assert. Either constrain the helpers on `TemplatePickerNode[T]` so the constraint earns its keep, or drop the constraint and the three asserts.

model/gurps/template_picker.go
```diff
@@ -62,3 +71,36 @@ func (t TemplatePicker) Hash(h hash.Hash) {
...
 }
+
+// HasTemplatePickerData returns true if any node or their child has non-zero template picker data
+func HasTemplatePickerData[T Node[T]](nodes ...T) bool {
+	for _, node := range nodes {
+		if !node.Container() {
+			continue
+		}
+		if tpp, ok := any(node).(TemplatePickerProvider); ok {
+			if _, data := tpp.TemplatePickerData(); data != nil && !data.IsZero() {
```
richardwilkes: No cases of `.TemplatePickerData()` return nil, as they all return the address of the embedded field. No need to check for nil here.

model/gurps/template_picker.go
```diff
@@ -62,3 +71,36 @@ func (t TemplatePicker) Hash(h hash.Hash) {
 		t.Qualifier.Hash(h)
 	}
 }
+
+// HasTemplatePickerData returns true if any node or their child has non-zero template picker data
+func HasTemplatePickerData[T Node[T]](nodes ...T) bool {
```
richardwilkes, line 76: From claude:

`model/gurps/template_picker.go:76-105` (nit): both helpers hand-roll a recursive tree walk that `gurps.Traverse` (`traverse.go:23`) already provides and that the merge code this PR moves (`mergePoints`) already uses. A callback returning `true` gives the early-exit "has" form; one returning `false` gives the "clear" form.

ux/template_merge.go
```diff
@@ -0,0 +1,202 @@
...
+func sameTechLevel(a, b gurps.TechLevelProvider) bool {
+	if a.RequiresTL() != b.RequiresTL() {
+		return false
+	}
+	return !a.RequiresTL() || a.TL() == b.TL()
+}
+
+// MergeableNode is the set of row types whose identical rows are merged by folding their points together: skills and
+// spells.
+type MergeableNode[T gurps.Node[T]] interface {
```
richardwilkes: This was made public, yet all of its uses are still on private functions/methods, so it should either remain private, or those uses should be exposed. I'd probably err on the side of keeping it private.

model/gurps/template_picker.go
```diff
@@ -62,3 +71,36 @@ func (t TemplatePicker) Hash(h hash.Hash) {
...
+		}
+		if HasTemplatePickerData(node.NodeChildren()...) {
+			return true
+		}
+	}
+	return false
+}
+
+// ClearTemplatePickerData removes the template picker data from the nodes and their children
+func ClearTemplatePickerData[T Node[T]](nodes ...T) {
```
richardwilkes, line 93: From claude:

`model/gurps/template_picker.go:94-105`: `ClearTemplatePickerData` zeroes a field that lives in the library-sync struct (`TraitContainerSyncData`, `SkillContainerOnlySyncData`, `SpellContainerOnlySyncData`) without touching the node's `Source`. `ApplySyncData` copies that whole struct back (`trait.go:1065`), so the next sync-with-library silently restores every cleared picker, and since `TemplatePicker` feeds `Hash` (`trait.go:1101`, `skill.go:1436`, `spell.go:1120`) the row reads as modified against its source until then. Any future caller needs to clear the source too (or the helper should), and the doc comment should say which.

---

Normally, I'd disagree that the Source should be cleared, but since this is just turning a template picker group into a normal group, that's probably the right thing to do.

model/gurps/template_picker.go
LeeSaferite: Yes, this is transitory code and will be used by upcoming PRs. I split it to make it easier. The assert has value on it's own, even if the helper functions aren't used right away. The code that would get the most use from this constraint cannot be switched just yet, but **will** be switched soon.

model/gurps/template_picker.go
LeeSaferite: While I agree that none of the current implementations return a nil, the interface *could* return a nil and the nil check was just safety. I can remove it if you prefer.

richardwilkes: It doesn't hurt anything. Go ahead and keep it.

model/gurps/template_picker.go
LeeSaferite: Funny thing is, I have another PR on it's way that's clearing source when you enable template choices on a container.

LeeSaferite: The reality here is that the only place that should ever *have* template picker values is *also* not a valid source ref location and, as such, the copied node will never have a sync source.

richardwilkes (approved):

